PR #49, opened 2015-01-04
Update to rustc 0.13.0-nightly

rolandshoemaker:
(hopefully) does what it says on the tin


huonw:
(I would merge, but the build seems to be failing because crates.io's regex is out of date, and I don't have the permissions publish the fixed version yet.)

alexcrichton:
Thanks!

PR #50, opened 2015-01-05
Add a higher level wrapper around precise_time_ns()

sfackler:
Since precise_time_ns() returns a number of nanoseconds since an
unspecified epoch, the set of operations that can be performed is a bit
limited. The epoch could theoretically be far enough in the past that
the value overflows during the runtime of the program, so the values
can't even necessarily be compared via the normal ordering operators.

The `PreciseTimestamp` type is an opaque wrapper around the precise
time, which only provides operations that actually make sense to use.
Right now that's only a method that turns two `PreciseTimestamps` into
the `Duration` of time between them. Since we care about overflow, we
*assume* that the `later` timestamp was created after the `self`
timestamp, and the result is positive no matter what (unless you waited
more than 293 years between measurements).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -253,6 +253,53 @@ pub fn precise_time_s() -> f64 {
     return (precise_time_ns() as f64) / 1000000000.;
 }
 
+/// An opaque structure representing a moment in time.
+///
+/// The only operation that can be performed on a `PreciseTime` is the
+/// calculation of the `Duration` of time that lies between them.
+///
+/// # Examples
+///
+/// Repeatedly call a function for 1 second:
+/// ```rust
+/// # fn do_some_work() {}
+/// let start = PreciseTime::now();
+///
+/// while start.to(PreciseTime::now()) < Duration::seconds(1) {
+///     do_some_work();
+/// }
+/// ```
+#[derive(Copy, Clone)]
+pub struct PreciseTime(u64);
+
+impl PreciseTime {
+    /// Returns a `PreciseTime` representing the current moment in time.
+    pub fn now() -> PreciseTime {
+        PreciseTime(precise_time_ns())
+    }
+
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -253,6 +253,55 @@ pub fn precise_time_s() -> f64 {
...
+        //
+        // We could deal with the overflow when casting to an i64, but all that
+        // gets us is the ability to handle intervals of up to 584 years, which
+        // seems not very useful :)
+        Duration::nanoseconds((later.0 - self.0) as i64)
+    }
+}
+
+/// Returns a `PreciseTimestamp` representing the current moment in time.
+pub fn precise_time() -> PreciseTimestamp {
```
alexcrichton: This construction seems a little odd for two reasons I think:

* The name `precise_time` and the returned name `PreciseTimespec` differ.
* This doesn't follow our constructor conventions of `Foo::new()`. It does align with `precise_time_ns`, but perhaps a more official struct should have a more conventional construction location?

In general this whole crate needs an overhaul though, so I'm certainly not opposed to incremental improvements! Curious what you think though.

sfackler: Updated

sfackler:
I went with `PreciseTime::now()` instead of `PreciseTime::new()` since it seems a bit clearer about what it's doing and still seems to be in line with naming conventions (e.g. `Thread::spawn`).

alexcrichton:
Cool!

sfackler:
One thing to note is that I'm not entirely confident that it behaves correctly in the presence of overflow. I somehow doubt that all three implementations of `precise_time_ns` overflow at 2^64 nanoseconds exactly.

We we may need to have OS-specific implementations of `PreciseTime` that keep the full set of information around or something. On the other hand, it may be the case that overflow doesn't actually happen in practice and we can ignore it, which would allow the API to expand a lot.

I'll poke around.

PR #54, opened 2015-01-07
Use ONCE.call_once instead of the old ONCE.doit

michaelcontento:
Changed in this PR https://github.com/rust-lang/rust/pull/20315


alexcrichton:
Thanks!

PR #55, opened 2015-01-08
Implement std::fmt::String for ParseError and TmFmt

akiss77:


akiss77:
An alternative solution got implemented and landed. Closing.

alexcrichton:
Ack sorry I overlooked this! I'll try to be more on top of this next time!

akiss77:
Np. The one that got landed looks better actually.

PR #59, opened 2015-01-16
fix broken documentation markup

fhartwig:
(I would have indented the code block one more column, but for some reason doing that causes rustdoc to crash.)


huonw:
Thanks!

> for some reason doing that causes rustdoc to crash.)

Indenting 4 spaces makes the whole thing considered  code block, including the \`\`\`'s, which aren't valid Rust so rustdoc (which tries to tokenise it to syntax highlight, and compile it to run tests) will be unhappy. In any cas, I think the indentation is unnecessary, it's just putting the blank line between the paragraph and the start of the codeblock (the ```rust) that fixes this problem.

Also, according to travis, making it considered code properly is causing it to fail tests (i.e. it wasn't valid code), the fix would be to import the two types from time `use time::{PreciseTime, Duration};` and to define the `do_some_work` function, even just `fn do_some_work() { /* ... */ }` at the  end of the code block is good enough.

fhartwig:
Thanks for the review! I've undone the indentation and added the necessary imports, the tests pass now. Should I squash this into a single commit?

huonw:
Yes, squashing would be useful if you can.

(Sorry for the delay @fhartwig, if I'm ever so slow again feel free to comment as a reminder.)

fhartwig:
Squashed and rebased. No worries about the delay, this wasn't particularly urgent (-:

PR #61, opened 2015-01-18
Add a more robust monotonic time wrapper

sfackler:
It looks like overflow isn't really a thing worth worrying about.
Windows guarantees 100 years of uptime before overflow of
QueryPerformanceCounter, and the OSX and Linux timers appear to be based
off of boot time as well.

Storing the time as the source data type makes the implementation more
complex, but avoids precision loss or shifts due to conversions to the
greatest extent possible.


sfackler:
I considered merging this with the existing crappy PreciseTime thing I added a couple weeks ago. Thoughts?

alexcrichton:
I suppose I don't quite know what the concept of a "precise time", "steady time", or all other kinds of time are. Would you feel up to the task of documenting why we have each kind of time? It'd be nice to at least know what the "steady" is in SteadyTime. I could also envision this replacing `PreciseTime` if necessary (this seems to be a clone of it, right?)

sfackler:
Sure. The term "steady" is taken from C++'s chrono package, meaning a source of "time units" that are monotonically increasing at a fixed rate, i.e. not susceptible to shifts or skewing: http://www.cplusplus.com/reference/chrono/steady_clock/. It has a separate "high resolution clock" that's not guaranteed to be monotonic or fixed-rate.

I could go either way on resolving the `PreciseTime`/`SteadyTime` issue. I could rename this new thing to `PreciseTime` and move `PreciseTime`'s existing method over and deprecate it, or deprecate `PreciseTime` all together in favor of the name `SteadyTime`. I'm kind of leaning towards the latter since it'll mach more closely to C++.

In any case, it'd probably make sense to deprecate `precise_time_ns` and `precise_time_s` as well.

sfackler:
Updated.

alexcrichton:
Thanks!

PR #62, opened 2015-01-21
fixes build for mac platforms

shadowmint:
Fixes the build for mac platforms. 

Currently errors with:

    src/lib.rs:249:20: 249:45 error: call to unsafe function requires unsafe function or block [E0133]
    src/lib.rs:249         let time = imp::mach_absolute_time();
                                      ^~~~~~~~~~~~~~~~~~~~~~~~~
    Could not compile `time`.


alexcrichton:
Thanks!

PR #68, opened 2015-02-06
Remove inclusion of tchar.h

bryal:
The dependency on tchar.h prevented build on windows for me. Removal of the inclusion does not appear to cause any problems. Unless this will cause other problems, I'd like to have it removed.


alexcrichton:
Thanks!

PR #73, opened 2015-02-26
Bump to 0.1.19 and dependency on rustc-serialize

fsommar:


fsommar:
It's in a pretty awkward state now that most libraries are starting to migrate to the new IO crate. On second thought, it might not be the best idea to upgrade the dependency on rustc-serialize at this point in time?

alexcrichton:
I'm fine with this plan, thanks @fsommar!

PR #74, opened 2015-03-01
Attribute is no longer needed as #[deny(stable_features)] is on by default

faineance:
Fixes build https://travis-ci.org/rust-lang/time/builds/52219903


alexcrichton:
Thanks!

PR #77, opened 2015-03-08
Support for subtracting two `Tm`s, yielding a Duration.

drewcrawford:
I release this patch under the MIT license.


alexcrichton:
Thanks!

PR #83, opened 2015-03-29
enable build on openbsd/bitrig

semarie:
sync with rust (src/libstd/sys/unix/time.rs)

openbsd/bitrig provides `clock_gettime` via C library.


alexcrichton:
Thanks!

PR #86, opened 2015-04-03
Fix the conversion from filetime to ns

LorenVS:
I might be wrong, but the current code does not seem to line up with the comment on filetimes. Hopefully this is the correct fix

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -217,7 +217,7 @@ pub fn get_time() -> Timespec {
         // http://support.microsoft.com/kb/167296/en-us
         let us_since_1601 = (((time.dwHighDateTime as u64) << 32) |
                              ((time.dwLowDateTime  as u64) <<  0)) / 10;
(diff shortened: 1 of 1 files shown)


LorenVS:
Turns out my initial guess here was wrong. The code is actually functioning correctly, but the naming of many of the variables is suspect. By dividing by 10 (the original code), a conversion was being done from 100ns buckets to microseconds, not back to nanoseconds. A number of variables that were named ns_ were actually storing quantities in microseconds.

This PR no longer affects functionality, but solves some naming issues with these variables

LorenVS:
This begs the question, since Timespec uses nanosecond precision, whether throwing out the 100ns precision is correct.

alexcrichton:
Aha! I looked into this a bit as well and I thought the number of 0s on the end of that constant looked suspicious... Thanks for this!

alexcrichton:
I'd also be fine actually using the ns precision if possible :)

PR #87, opened 2015-04-03
Do not convert to and from timespec when calling at_utc on a utc Tm

### LINKED ISSUE #66: libtime::Tm::to_utc() is slow even if already utc
<a href="https://github.com/seanmonstar"></img></a> **Issue by [seanmonstar](https://github.com/seanmonstar)**
_Friday Nov 07, 2014 at 01:29 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/18725_

_This issue was labelled with:  in the Rust repository_

----

`Tm.to_utc()` calls out to the runtime fn `rust_timegm` twice, once to convert to a Timespec, and a second time to convert the timespec to a new Tm with offset of 0. If a `Tm` is already UTC, this should skip both those calls.

LorenVS:
Fixes #66 , based on the suggestion by @zslayton

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -769,7 +769,10 @@ impl Tm {
 
     /// Convert time to the UTC
     pub fn to_utc(&self) -> Tm {
-        at_utc(self.to_timespec())
+        match self.tm_utcoff {
+            0 => *self,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #88, opened 2015-04-03
strptime now supports leading zeroes

### LINKED ISSUE #25: strptime() not handling optionality of leading zeroes
For many format specifiers, leading zeroes are supposed to be optional:  http://pubs.opengroup.org/onlinepubs/7908799/xsh/strptime.html

I have only tested %m.

```toml
[package]
name = "test"
version = "0.0.0"
authors = [ "mike@mikedilger.com" ]

[[bin]]
name = "test"
path = "main.rs"

[dependencies]
time = "*"
```

```rust
extern crate time;

use time::strptime;

fn main() {
    // %m is the month number [1,12]; leading zeros are permitted but not required.
    let format = "%d/%m/%Y".into_string();
    let dates = [ "24/05/2016", "24/5/2016" ];

    for date in dates.iter() {
        match strptime(*date, format.as_slice() ) {
            Ok(d) => println!("{} is valid {}, result shows as {}", date, format, d),
            Err(e) => println!("{} is invalid {} -- {}", date, format, e),
        }
    }
}
```

```
24/05/2016 is valid %d/%m/%Y, result shows as Tm { tm_sec: 0, tm_min: 0, tm_hour: 0, tm_mday: 24, tm_mon: 4, tm_year: 116, tm_wday: 0, tm_yday: 0, tm_isdst: 0, tm_utcoff: 0, tm_nsec: 0 }
24/5/2016 is invalid %d/%m/%Y -- Invalid month.
```

LorenVS:
Fixes #25 . Does anyone have any suggestion on how to keep the check on line lib.rs:1184? I can't think of a way that we can make such a check work if there are multiple ways of encoding the same possible date string.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1178,7 +1178,6 @@ mod tests {
             match strptime(s, format) {
               Ok(_) => {
                 tm.strftime(format).unwrap().to_string() == s.to_string()
-                true
               },
               Err(e) => panic!("{:?},  s={:?}, format={:?}", e, s, format)
             }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1176,8 +1176,13 @@ mod tests {
 
         fn test(s: &str, format: &str) -> bool {
             match strptime(s, format) {
-              Ok(ref tm) => {
-                tm.strftime(format).unwrap().to_string() == s.to_string()
+              Ok(_) => {
+                // this condition is nice, but since numeric tokens can
+                // be prefixed by either ' ' or '0', a specific date
+                // can be serialized back to multiple date strings, so
+                // we can't rely on this test
+                // tm.strftime(format).unwrap().to_string() == s.to_string()
+                true
               },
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1176,8 +1176,13 @@ mod tests {
...
             match strptime(s, format) {
-              Ok(ref tm) => {
-                tm.strftime(format).unwrap().to_string() == s.to_string()
+              Ok(_) => {
+                // this condition is nice, but since numeric tokens can
+                // be prefixed by either ' ' or '0', a specific date
+                // can be serialized back to multiple date strings, so
+                // we can't rely on this test
+                // tm.strftime(format).unwrap().to_string() == s.to_string()
+                true
```
alexcrichton: This seems like a pretty important assertion here actually, would it be possible to keep this but perhaps have a specialized test for those which can't pass this check?

LorenVS:
I don't have access to a machine to test this on right now, but hopefully this is what you had in mind

alexcrichton:
Thanks!

PR #89, opened 2015-04-03
Allow colon in timezone specification

### LINKED ISSUE #20: strptime and time zones
strptime is not handling time zones correctly. For example, if you use the format string "%FT%T%z" with ISO 8601 strings like
1999-05-31T13:20:00-04:00
and
1999-05-31T13:20:00-05:00
you get identical results back.

The code on line 510 seems seems to match the time zone offset but it only does something when the offset if zero.

*Imported from https://github.com/rust-lang/rust/issues/2637*

LorenVS:
Fixes #20 

This PR allows (but does not insist) on having colons separating the hours and minutes portion of a timezone specifier.

This PR depends on the strptime leading zeroes support. If that PR doesn't get merged in I can update this one to not depend on it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1304,9 +1304,13 @@ mod tests {
         assert!(test_oneway("3",  "%S"));
 
         assert!(strptime("-0000", "%z").unwrap().tm_utcoff == 0);
+        assert!(strptime("-00:00", "%z").unwrap().tm_utcoff == 0);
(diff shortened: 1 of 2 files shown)


LorenVS:
Sorry for the PR spam, I won't have a rust environment available to me until tonight, and apparently my Rust foo is still pretty weak :(

alexcrichton:
No problem! I've tested locally and it all works just fine!

PR #90, opened 2015-04-03
Allow parsing epoch seconds

### LINKED ISSUE #58: strptime add support to convert epoch time string
Hi,

i am trying to convert an epoch time in seconds to a time::Tm.
It seems that "%s" format is not recognized, I get an InvalidFormatSpecifier('s') in my example below.

```
extern crate time;
use time::strptime;

fn main() {
    let format = "%s";
    let date = "1236905626";

        match strptime(date, format ) {
            Ok(d) => println!("{} is valid {}, result shows as {:?}", date,  format, d),
            Err(e) => println!("{:?}", e),
        }
}
```

rustc --version:
```
rustc 1.0.0-nightly (3d0d9bb6f 2015-01-12 22:56:20 +0000)
```
my Cargo.toml 

```
[package]
name = "test"
version = "0.0.0"
authors = [ "lucdew@lucdew.com" ]

[[bin]]
name = "test"
path = "main.rs"

[dependencies]
time = "*"

```

LorenVS:
Fixes #58 

This PR adds the ability to parse %s format strings. Its a little interesting since %s basically dictates every possible value of the Tm struct, including utc offset, so I just overwrite the working Tm instance with a new Tm from the parsed %s field. Hopefully that is correct!

Once again, this PR depends on the leading zeroes PR. If that doesn't get merged I can redo this one to be independent. Just ignore the first commit if you are looking for changes that are relevant to this PR

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,8 +35,8 @@ use std::io;
 use self::ParseError::{InvalidDay, InvalidDayOfMonth, InvalidDayOfWeek,
                        InvalidDayOfYear, InvalidFormatSpecifier, InvalidHour,
                        InvalidMinute, InvalidMonth, InvalidSecond, InvalidTime,
-                       InvalidYear, InvalidZoneOffset, MissingFormatConverter,
-                       UnexpectedCharacter};
+                       InvalidYear, InvalidZoneOffset, InvalidSecondsSinceEpoch,
+                       MissingFormatConverter, UnexpectedCharacter};
 
 pub use parse::strptime;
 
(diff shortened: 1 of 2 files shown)


LorenVS:
Anything else outstanding here?

src/lib.rs
```diff
@@ -1176,8 +1178,13 @@ mod tests {
...
             match strptime(s, format) {
-              Ok(ref tm) => {
-                tm.strftime(format).unwrap().to_string() == s.to_string()
+              Ok(_) => {
+                // this condition is nice, but since numeric tokens can
+                // be prefixed by either ' ' or '0', a specific date
+                // can be serialized back to multiple date strings, so
+                // we can't rely on this test
+                // tm.strftime(format).unwrap().to_string() == s.to_string()
+                true
```
alexcrichton: Did this sneak back in by accident? (I think this was removed in a previous commit)

LorenVS:
Rebasing seems to have fixed the diff

alexcrichton:
Thanks!

PR #97, opened 2015-05-12
implement FromStr Trait for Tm

ledbettj:
Current implementation only accepts a few formats, but easy enough to add new ones.

Useful, for example, for automatically parsing Time objects from command line arguments via [argparse](https://github.com/tailhook/rust-argparse).

Let me know if there's any feedback, changes, or discussion needed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,6 +44,7 @@ use std::cmp::Ordering;
 use std::fmt;
 use std::ops::{Add, Sub};
 use std::io;
+use std::str::FromStr;
 
 #[cfg(feature = "std-duration")]      pub use std::time::Duration;
 #[cfg(not(feature = "std-duration"))] pub use duration::Duration;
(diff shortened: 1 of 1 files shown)


alexcrichton:
I'm a little worried about this being a little too "magical" as there's not really a canonical way to go from a time to a string and back. Picking a few common date formats is good for ease-of-use but can often end up leading to surprises later on down the road.

ledbettj:
That's a valid concern that I didn't really consider.  If `FromStr` is supposed to parse from exactly one canonical format, this behavior is definitely wrong.  Perhaps if it 'only' accepted ISO 8601 formatted values?

alexcrichton:
Yeah I think that parsing only one specific format would be more acceptable, but I suspect that we may not want to decree the "one true format" of `Tm` via the `FromStr` implementation, however. Each application may have its own "canonical date format" different from all the others.

ledbettj:
Sounds good.  Closing!

PR #99, opened 2015-06-01
Clarification of how nsec are used in Timespec

lalala:
Clarification of how nsec are used in Timespec to represent a time offset. The documentation on this is ambiguous as one could infer a different functioning.


alexcrichton:
Thanks!

PR #100, opened 2015-06-08
parse TZ offset of Z as UTC

nerdrew:
http://en.wikipedia.org/wiki/ISO_8601#UTC

`2015-06-08T05:53:01Z` has a valid ISO 8601 timezone.


alexcrichton:
Thanks!

PR #103, opened 2015-06-23
remove reliance on C code, move most into sys.rs

### LINKED ISSUE #78: Remove C dependencies
I recently had to compile a Rust project with time as a dependency and I went through a lot of trouble having to get mingw64 on my 64 bit machine and messing around with paths to make sure that's the version that's used and not the 32 bit version. (I have 64 bit Rust and 32 bit C mingw)

Is it possible to remove the C dependencies so this project uses only Rust? That would make compiling projects with time as a dependency much easier on Windows.

seanmonstar:
This builds on the 2 machines I have access to:

- Windows 7 x86_64
- `Linux grunt 4.0.4-x86_64-linode57 #1 SMP Thu May 21 11:01:47 EDT 2015 x86_64 x86_64 x86_64 GNU/Linux`

I know there were some tiny differences in time_helpers.c, such as for native client, that I can't test, nor do I specifically know how to port it into Rust.

Fixes #78


retep998:
Note this does not pass tests because the tests set and assume a specific timezone but the winapi time code only respects the global system timezone.

alexcrichton:
Thanks! Looks like there's some travis failures, however.

seanmonstar:
It passed!

seanmonstar:
Granted, this hasn't been tested on Mac, NaCl, or other versions of unix...

src/sys.rs
```diff
@@ -0,0 +1,425 @@
...
+        tm_mon: c_int,
+        tm_year: c_int,
+        tm_wday: c_int,
+        tm_yday: c_int,
+        tm_isdst: c_int,
+        tm_gmtoff: c_long,
+        tm_zone: *const c_char,
+    }
+
+    impl Default for tm {
```
alexcrichton: I'd prefer to not implement traits for `#[repr(C)]` types, can this just use `mem::zeroed` below?

src/sys.rs
```diff
@@ -0,0 +1,425 @@
...
+                tm_wday: 0,
+                tm_yday: 0,
+                tm_isdst: 0,
+                tm_gmtoff: 0,
+                tm_zone: 0 as *const c_char
+            }
+        }
+    }
+
+    impl<'a> From<&'a Tm> for tm {
```
alexcrichton: Same from this as above, I think this should be a function. Also, the old C version did not set gmtoff, which may be significant.

src/sys.rs
```diff
@@ -0,0 +1,425 @@
...
+                tm_wday: tm.tm_wday,
+                tm_yday: tm.tm_yday,
+                tm_isdst: tm.tm_isdst,
+                tm_gmtoff: tm.tm_utcoff as c_long,
+                .. tm::default()
+            }
+        }
+    }
+
+    fn tm_to_rust_tm(tm: &tm, rust_tm: &mut Tm) {
```
alexcrichton: The old C version also takes a utcoff and nsec set of arguments which I believe are needed here.

src/sys.rs
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    #[cfg(all(not(target_os = "macos"), not(target_os = "ios")))]
+    pub fn get_precise_ns() -> u64 {
+        let mut ts = libc::timespec { tv_sec: 0, tv_nsec: 0 };
+        unsafe {
+            unix::clock_gettime(libc::CLOCK_MONOTONIC, &mut ts);
+        }
+        (ts.tv_sec as u64) * 1000000000 + (ts.tv_nsec as u64)
+    }
```
alexcrichton: Can these `#[cfg]` functions move within the module they should be defined in? It would cut down on the `#[cfg]` and avoid exposing the internals as `pub`

src/sys.rs
```diff
@@ -0,0 +1,425 @@
...
+    struct SystemTime {
+        wYear: WORD,
+        wMonth: WORD,
+        wDayOfWeek: WORD,
+        wDay: WORD,
+        wHour: WORD,
+        wMinute: WORD,
+        wSecond: WORD,
+        wMilliseconds: WORD,
+    }
```
alexcrichton: Can this use winapi instead of defining everything here?

alexcrichton:
This is quite a critical crate and the tests I don't believe are anywhere near exhaustive, can you try this update out on some local crates which depend on it and make sure the tests pass? I will do so locally as well, but porting away from C is traditionally fraught with errors.

src/sys.rs
seanmonstar: The `nsec` is adjusted in lib.rs, after the call to sys. The utcoff is adjusted adjust outside this function, [here](https://github.com/rust-lang/time/pull/103/files#diff-3b6a6b2a9f8fe54eba20f71df5805035R91) and [here](https://github.com/rust-lang/time/pull/103/files#diff-3b6a6b2a9f8fe54eba20f71df5805035R100)

src/sys.rs
seanmonstar: I could use `mem::zeroed`. What's wrong with implementing traits on `repr(C)`? Does the method get exposed on the C side?

src/sys.rs
seanmonstar: Ah, I could move these methods into `unix` and `mac` modules, if that's what you mean.

src/sys.rs
seanmonstar: Do you mean [this winapi](https://crates.io/crates/winapi)? We can. The types and functions used in this module seem to be [unimplemented](https://github.com/retep998/winapi-rs/blob/master/lib/kernel32-sys/src/lib.rs#L369).

seanmonstar:
I agree that this initial work likely breaks on some system, as there's plenty of `#define` magic that was happening in `time_helpers.c`. Is it possible to use the rust buildbot infrastructure to help test on more machines?

src/sys.rs
alexcrichton: It's not quite idiomatic as for example all the types in liblibc don't implement this.

alexcrichton: Could you consolidate the logic into this function? It's a little unfortunate having it spread out over so many different areas.

src/sys.rs
alexcrichton: Could you consolidate the logic into this function? It's a little unfortunate having it spread out over so many different areas.

src/sys.rs
alexcrichton: I do indeed! I suspect that @retep998 would be willing to accept patches to upstream these headers into winapi and one of the `*-sys` crates.

retep998: If I don't see a pull request for it by the time I wake up, I'll just add that stuff myself.

src/sys.rs
seanmonstar: I've added the `utcoff` to this function. I left `nsec` in lib.rs, since it's not useful to any of the ffi functions, and instead of having to set `tm.tm_nsec` in 4 different functions (unix and windows), it can just be in 2 functions (`at` and `at_utc`).

src/sys.rs
seanmonstar: Converted to function, and stopped setting gmtoff.

src/sys.rs
seanmonstar: Ok.

src/lib.rs
```diff
@@ -609,19 +468,22 @@ mod steady {
...
                 as libc::LARGE_INTEGER;
             self
         }
     }
 }
-
+#[cfg(not(windows))]
 pub fn tzset() {
-    #[cfg(windows)] extern { #[link_name = "_tzset"] fn tzset(); }
     #[cfg(not(windows))] extern { fn tzset(); }
```
retep998: You can remove this extra `#[cfg]` since the function itself has the same one.

seanmonstar: Good catch.

seanmonstar:
I also noticed that `SteadyTime` code could be reduced to a single implementation, since they all just used `precise_time_ns` anyway. I included that in a second commit.

src/sys.rs
```diff
@@ -0,0 +1,349 @@
...
+#[allow(non_snake_case)]
+mod inner {
+    use ::Tm;
+    use std::mem;
+    use std::sync::{Once, ONCE_INIT};
+
+    use winapi::*;
+    use kernel32::*;
+
+    fn frequency() -> libc::LARGE_INTEGER {
```
retep998: Just `LARGE_INTEGER`. It is provided by `winapi`, no need to refer to `libc` which isn't even used in this module.

src/sys.rs
```diff
@@ -0,0 +1,349 @@
...
+        (((t - HECTONANOSEC_TO_UNIX_EPOCH) % HECTONANOSECS_IN_SEC) * 100) as i32
+    }
+
+    fn file_time_to_time(ft: &FILETIME) -> i64 {
+        let t = ((ft.dwHighDateTime as u64) << 32) | (ft.dwLowDateTime as u64);
+        ((t - HECTONANOSEC_TO_UNIX_EPOCH) / HECTONANOSECS_IN_SEC) as i64
+    }
+
+    fn tm_to_system_time(tm: &Tm) -> SYSTEMTIME {
+        let mut sys = unsafe { mem::zeroed() };
```
retep998: let mut sys: SYSTEMTIME   = unsafe { mem::zeroed() };

src/sys.rs
```diff
@@ -0,0 +1,349 @@
...
+        system_time_to_tm(&out, tm);
+        tm.tm_utcoff = 0;
+    }
+
+    pub fn time_to_local_tm(sec: i64, tm: &mut Tm) {
+        let mut out = unsafe { mem::zeroed() };
+        let tz = time_zone(None);
+        let ft = time_to_file_time(sec);
+        unsafe {
+            let mut utc = unsafe { mem::zeroed() };
```
retep998: This `unsafe` isn't needed.

sfackler:
SteadyTime was structured the way it was to maintain the native system time types for as long as possible to make sure it stays consistent in the face of any overflow that could occur (I believe only windows provides any guarantees about when the monotonic time source overflows).

alexcrichton:
I've added a few commits on top of this in a `tmp` branch of this repo to test things out, but I'll comment with a few other questions inline as well:

* error handling on unix: https://github.com/rust-lang/time/commit/9eaef66243a9cb6b78f3d5cd8e5c9439cbecb3e9
* error handling on windows: https://github.com/rust-lang/time/commit/37da3861ae1097a71add1988f26a0a0524e56b41
* clarify seconds are unix seconds on windows: https://github.com/rust-lang/time/commit/3cd19dedec1328f7822ce8f06273ca0c2f71aa2f

src/sys.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub fn get_precise_ns() -> u64 {
+        let mut ticks = 0;
+        unsafe {
+            assert!(QueryPerformanceCounter(&mut ticks) == 1);
+        }
+        mul_div_i64(ticks as i64, 1000000000, frequency() as i64) as u64
+
+    }
+
+    pub fn time_zone(bias: Option<i32>) -> *const TIME_ZONE_INFORMATION {
```
alexcrichton: I'm not sure I quite understand what's going on in this function, is it trying to emulate some unix primitive? There are a number of calls in this file to `time_zone`, in addition to some calls in some tests, but this is guarded with a `Once` so I'm not sure what effect they all have.

seanmonstar: Ah, I should have left comments explaining what this is. This came about with help from @retep998. 

The previous environment variable strategy doesn't work with Windows' time functions; you can pass a optional time zone info instead. Passing a null pointer means it will default to the local time. Setting the time zone is only ever used in the test suite, so it's not exposed publicly. The other calls to it are to fetch a possibly set time zone, so the tests can pass.

alexcrichton: Hm ok! I'll add some comments, but I may remove this altogether and just update the tests.

alexcrichton:
OK, touched up a few more things and pushed to master, thanks again @seanmonstar!

seanmonstar:
Excited and scared, as I'm not sure if much of `#define` magic that was in the C file was ported properly/tested.

alexcrichton:
I verified as much as I could manually, and I think it all checks out at least, but I suspect various platforms like Android may require some follow-up patches. Regardless this is certainly a great step forward for this dependency!

I've published 0.1.27 with this patch.

sfackler:
@alexcrichton  Are we intentionally going forward with the SteadyTime changes or what?

alexcrichton:
Ah I missed that, I will revert.

seanmonstar:
So that I understand, the point of the `SteadyTime` implementations is to do the math to convert to nanoseconds only when desired, but otherwise leaving at the highest resolution provided by the OS?

sfackler:
It wants to preserve both the resolution and the range.

PR #104, opened 2015-06-25
Make Tm and Timespec Hash

brandonson:


brandonson:
I can't see any reason for them not to be - if I'm missing something, let me know.

alexcrichton:
Thanks!

PR #105, opened 2015-06-27
Added ToJson implementation for Timespec.

kamarkiewicz:
Hi! I needed impl of ToJson for Timespec so I code it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ Utilities for working with time-related functions in Rust.
 [dependencies]
 libc = "0.1"
 rustc-serialize = { version = "0.3", optional = true }
+postgres = { version = "*", optional = true }
+byteorder = "*" # required by postgres
 
 [build-dependencies]
 gcc = "0.3"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -15,6 +15,8 @@ Utilities for working with time-related functions in Rust.
 [dependencies]
 libc = "0.1"
 rustc-serialize = { version = "0.3", optional = true }
+postgres = { version = "*", optional = true }
```
alexcrichton: I believe that [postgres is responsible](https://github.com/sfackler/rust-postgres/blob/master/Cargo.toml#L36) for this functionality instead of time itself, so I don't think this should be added just yet.

alexcrichton:
I'm a little uncertain about this because there's not a clear representation of a time in JSON, and the representation emitted here is highly unlikely to be any standard representation as well.

kamarkiewicz:
Ugh, I screw sth with this pull request... I meant to pass only commit #d4e9148.
You're right, but following your way there shouldn't be RustcEncodable and RustcDecodable as well. Am I wrong?

kamarkiewicz:
I will pass pull request again, but this time the right way...

PR #106, opened 2015-06-27
ToJson implementation for Timespec

kamarkiewicz:
Why RustcEncodable and RustcEncodable are derived for Timespec and there is no ToJson trivial implementation?
I've messed up there #105 so I've made a new request.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,8 @@ use std::cmp::Ordering;
 use std::fmt;
 use std::ops::{Add, Sub};
 use std::io;
+#[cfg(feature = "rustc-serialize")]
+use rustc_serialize::json::{self, Json, ToJson};
 
 pub use duration::Duration;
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -208,6 +210,13 @@ impl Sub<Timespec> for Timespec {
     }
 }
 
+#[cfg(feature = "rustc-serialize")]
+impl ToJson for Timespec {
+    fn to_json(&self) -> Json {
+        Json::String(json::encode(&self).unwrap())
```
sfackler: This is going to end up creating things like

```json
{
    "my_time": "{\"sec\": 894582834, \"nsec\": 1200}"
}
```

which is probably not what you want.

alexcrichton:
I feel that this isn't really the representation that would be expected for a time in JSON, so I'm not sure that this trait should necessarily be implemented at all?

huonw:
I think the conventional representation is the one javascript uses, e.g. `"2015-07-08T17:30:50.906Z"`.

PR #110, opened 2015-07-06
Arithmetic overflow in SteadyTime subtraction on OS X / iOS

bmastenbrook:
Subtracting two SteadyTime values fails on OS X or iOS if the resulting Duration would be negative. Casting before the subtraction fixes the issue. I don't believe the cast itself can ever fail, as the value is based on a time elapsed since system boot and has a resolution of at most (least?) one nanosecond.


alexcrichton:
Can you add a test for this as well?

bmastenbrook:
Sure thing.

huonw:
Thanks!

PR #113, opened 2015-07-29
Make it compatible with Android.

overminder:
Android doesn't have `timegm` (it has `timegm64` though). I have thus added
an implementation for `timegm` based on `timegm64`.

All the test cases passed on my Android 4.2.2.


alexcrichton:
Thanks!

PR #114, opened 2015-08-02
Fix NaCl: Newlib lacks `timegm`, so use a crude workaround instead.

DiamondLovesYou:


alexcrichton:
This seems like it may be wrong in the face of threads, and it also introduces an environment variable when one may not have existed before. Are there threads on NaCl?

DiamondLovesYou:
@alexcrichton
> This seems like it may be wrong in the face of threads...

Yes, it is thread-unsafe. But on glibc `timegm` isn't thread safe, it's "MT-Safe env locale", so this patch is still technically correct.

> ...and it also introduces an environment variable when one may not have existed before.

Ah, I thought `std::env` lacked this functionality (I looked for `unset_var` instead of `remove_var` originally). Fixed.

> Are there threads on NaCl?

Indeed.

alexcrichton:
Ok, r=me with `var::remove` instead of setting it to an empty string.

DiamondLovesYou:
Oops forgot to push. Sorry about that.

PR #115, opened 2015-08-04
Added a short mention about the format string syntax used.

golddranks:
It struck me odd that this wasn't actually mentioned anywhere at the doc. Added a mention.


alexcrichton:
Thanks!

PR #118, opened 2015-08-20
Fixes timegm64 runtime linkage error on aarch64 Android.

mark-buer:
The symbol `timegm64` appears to be missing from the aarch64 android runtime, resulting in runtime linkage errors such as:

```text
java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "timegm64" referenced by ...
```

`__LP64__` is defined for aarch64, meaning the `timegm64` hacks are disabled. The [`time64.h`](https://android.googlesource.com/platform/development/+/797351fd3bbb8fe517afafdd5095fd740387e7a4/ndk/platforms/android-L/include/time64.h) header shows this to be the case.

The patch fixes the runtime linkage issue by only using the `timegm64` hacks for non-aarch64 Androids.


alexcrichton:
Thanks!

PR #119, opened 2015-09-10
Apply daylight savings on Windows

dten:
Apply DST info based off of the TIME_ZONE_INFORMATION for the relevant year using info from
https://msdn.microsoft.com/en-us/library/windows/desktop/ms725481(v=vs.85).aspx

Also don't crash for time before 1970 (pretty reasonable request)

### DIFF
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -214,6 +214,14 @@ mod inner {
         TzReset
     }
 
+    #[cfg(test)]
+    pub fn set_london_with_dst_time_zone() -> TzReset {
+        use std::env;
+        env::set_var("TZ", "Europe/London");
+        ::tzset();
+        TzReset
+    }
+
     #[cfg(all(not(target_os = "macos"), not(target_os = "ios")))]
     mod unix {
         use std::fmt;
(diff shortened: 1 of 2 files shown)


src/sys.rs
```diff
@@ -455,10 +455,39 @@ mod inner {
...
             system_time_to_tm(&local, tm);
-
+            tm.tm_utcoff = 0;
+            tm.tm_isdst = -1;
+            
             let mut tz = mem::zeroed();
-            GetTimeZoneInformation(&mut tz);
-            tm.tm_utcoff = -tz.Bias * 60;
+            let this_year = 1900 + tm.tm_year as u16;
+            let get_tz_info = GetTimeZoneInformationForYear(this_year, 0 as *mut _, &mut tz) != 0;
```
alexcrichton: Does this normally fail? The general pattern for these functions currently is to panic if they fail (e.g. the `call!` macro)

src/sys.rs
```diff
@@ -455,10 +455,39 @@ mod inner {
...
+
+            if get_tz_info {
+                tm.tm_isdst = 0;
+                tm.tm_utcoff -= tz.Bias * 60;
+
+                if tz.StandardDate.wMonth != 0 && tz.DaylightDate.wMonth != 0 {
```
alexcrichton: A few thoughts here:

* Could you add a link here to the MSDN page? Also, could you add a comment indicating that this is basically just trying to figure out if `tm` is within the range of DST and if so it uses the `DaylightBias` (otherwise `StandardBias` is used).
* The comparison operator may be relatively expensive (calls a few extra function calls) and deals with a lot of data, perhaps a more focused comparison could be used? (e.g. one that's entirely contained to just this module)
* It's a little confusing that `tm_utcoff` is set multiple times in this function, could the assignment to `tz.Bias * 60` happen in a fall-through branch that's only taken if `StandardBias` or `DaylightBias` isn't used?

alexcrichton:
Is there a way a test could be added for the DST aspects of this? It looks like the test added was for the underflow case?

dten:
Thanks for the comments; I will address them shortly

dten:
Hopefully that should do it... final solution is simpler

alexcrichton:
Thanks! Can you also squash the commits into one?

dten:
all squished

alexcrichton:
Thanks!

PR #120, opened 2015-09-22
Make kernel32-sys dependency version fuzzy for win32

vvuk:


alexcrichton:
Thanks for the PR! In Cargo, however, "0.1.2" actually means "semver compatible with 0.1.2" which also means "0.1.*", so it's already a bit of a fuzzy version. If 0.2 is ever released it could introduce breakage that this library isn't ready to handle, so I don't think we want to preemptively say we're compatible with that just yet.

As a result I'm closing for now.

PR #121, opened 2015-09-23
feat: add (de)serialization for time::Tm

compressed:
Similar to `Timespec` I had a need to serialize the `time::Tm` struct.


alexcrichton:
Thanks!

PR #131, opened 2016-02-01
Support heapsize

nox:


alexcrichton:
Unfortunately I'm not really sure what's going on here, but what is heapsize?

nox:
Oh yeah, that would be an useful detail, right? @Manishearth's heapsize is a crate that provides a trait HeapSizeOf that lets callers know how much memory consume some value. This is used in Servo for memory reporting and I am adding support for it in hyper which depends on time.

I believe this should be a feature of stdlib itself, but that's another story.

https://github.com/servo/heapsize

alexcrichton:
Ah ok, thanks for the explanation! I feel like this may be the wrong way to go about this, however, as we probably don't want to add this as a dependency to all crates in existence. Would it be possible for the heapsize crate itself to provide these sorts of implementations for the core types?

Manishearth:
> Would it be possible for the heapsize crate itself to provide these sorts of implementations for the core types?

Then heapsize has to add all crates as a dependency for it. It's not a clear cut choice.

nox:
> Would it be possible for the heapsize crate itself to provide these sorts of implementations for the core types?

This is no good, I have no way from outside to track the exact usage of complex things such as BTreeMap and LinkedList and HashMap, hence why I suggested that it should be a trait in Rust itself.

alexcrichton:
I think that we should set aside the should-this-be-in-std question for now, that can be debated later. For the `time` types specifically, I'm worried about having to add this feature basically everywhere to all rust-lang crates (which I would rather not do). Should all other derive-like traits be added as features to this crate? e.g. serde?

Overall I don't think that this is the right direction (adding a feature here) but rather adding a feature elsewhere to the heapsize crate.

nox:
@alexcrichton That works for `time` but not for other things with private fields etc.

alexcrichton:
Sure, but it's not really feasible to expect all crates to have an optional `heapsize` feature either. Nor is it really feasible to assume that most `#[derive]` like traits should be added to libstd.

nox:
Sure, but this sounds just as important as Debug to me, memory reporting that is.

oli-obk:
A similar topic has been [discussed on internals](https://internals.rust-lang.org/t/vague-proposal-extending-coherence-with-workspaces/3216?u=ker). 

I had a [suggestion](https://internals.rust-lang.org/t/vague-proposal-extending-coherence-with-workspaces/3216/7?u=ker) to use features, but have cargo enable them automatically depending on conditions.

And a related [rfc-issue](https://github.com/rust-lang/rfcs/issues/1553) about algorithmic generation of impls exists, too.

nox:
Actually I don't care about where the implementation lives, I care that HeapSizeOf should really be in Rust itself.

PR #133, opened 2016-03-30
Implement conversion to/from std::time::Duration

### LINKED ISSUE #128: Add `impl Into<std::time::Duration> for time::Duration` ?
I need a way to convert time::Duration to std::time::Duration, now that it's stable.

Maybe `From` would be useful too.

tailhook:
I'm not sure about exact error messages used. Also there is double check in `From<std::time::Duration>` which can be optimized, but I think this way it shows intent better. Hopefully compiler can optimize it for me.

Fixes #128

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -12,6 +12,7 @@
 
 use std::{fmt, i64};
 use std::ops::{Add, Sub, Mul, Div, Neg, FnOnce};
+use std::time::Duration as StdDuration;
 
 /// The number of nanoseconds in a microsecond.
 const NANOS_PER_MICRO: i32 = 1000;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm conventionally `from` and `into` are only intended for infallible conversions, so perhaps this could be done through inherent methods instead?

tailhook:
> Hm conventionally from and into are only intended for infallible conversions, so perhaps this could be done through inherent methods instead?

So are arithmetic operations. But we have an `arithmetic operation overflowed`.

Anyway, It's okay for me either way. How should inherent methods be called? `from/into` are okay? Should they panic or return an Option (or a Result)?

Another question, is I'm checking against `duration::MAX`. There are similar checks in `Duration::seconds` and similar constructors. But you can actually create a bigger duration by arithmetic (i.e. `MAX*2` works fine). Is the latter a bug? Or should we allow to create a Duration with up to a `Duration { secs: i64::MAX, ..}`  when converting?

alexcrichton:
Perhaps `Duration::from_std(std::time::Duration) -> Result<..>` and `Duration::to_std(&self) -> Result<std::time::Duration>` or something like that?

I'm not so sure on the bugs with MAX, I haven't looked at this library in awhile

alexcrichton:
Taken care of in https://github.com/rust-lang-deprecated/time/pull/134

PR #134, opened 2016-04-01
Implement conversion to/from std::time::Duration

### LINKED ISSUE #128: Add `impl Into<std::time::Duration> for time::Duration` ?

tailhook:
Fixes #128 and supersedes #133

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -11,7 +11,9 @@
 //! Temporal quantification
 
 use std::{fmt, i64};
+use std::error::Error;
 use std::ops::{Add, Sub, Mul, Div, Neg, FnOnce};
+use std::time::Duration as StdDuration;
 
 /// The number of nanoseconds in a microsecond.
 const NANOS_PER_MICRO: i32 = 1000;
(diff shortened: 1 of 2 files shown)


src/duration.rs
```diff
@@ -359,6 +391,27 @@ impl fmt::Display for Duration {
...
 }
 
+/// Represents error when converting `Duration` to/from a standard library
+/// implementation
+///
+/// The `std::time::Duration` supports a range from zero to `u64::MAX`
+/// *seconds*, while this module supports signed range of up to
+/// `i64::MAX` of *milliseconds*.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct OutOfRangeError;
```
alexcrichton: Could you make this `struct OutOfRangeError(())` so we can extend it in the future as well?

alexcrichton:
Ah I'll fix up the last nit manually, thanks!

PR #138, opened 2016-08-31
Add "no longer in active maintenance phase" notice to README

### LINKED ISSUE #136: Is this crate deprecated? Or isn't it? What to use instead?
The repo lives under the `rust-lang-deprecated` organisation. https://github.com/rust-lang/time redirects to this repo. Does that mean this crate is deprecated?

If this crate is indeed deprecated, what should I use as replacement?

nabijaczleweli:
Closes #136


alexcrichton:
Thanks!

PR #139, opened 2016-09-12
Add Duration::span_v() which also returns the closure's return value

nabijaczleweli:
I found myself needing to measure resultsome closures, this seemed like a simple fix.

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -139,6 +139,14 @@ impl Duration {
         Duration::nanoseconds((super::precise_time_ns() - before) as i64)
     }
 
+    /// Runs a closure, returning the duration of time it took to run the
+    /// closure and the closure's return value.
+    pub fn span_v<F, R>(f: F) -> (Duration, R) where F: FnOnce() -> R {
+        let before = super::precise_time_ns();
+        let f_res = f();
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Unfortunately though this crate is sort of in limbo and is essentially at the point where we're not accepting breaking changes right now. This is unfortunately a breaking change to the signature of the function :(

nabijaczleweli:
:sadfaec:
Maybe as another function (`span_r` or w/e, I'm bad at naming), though?

alexcrichton:
Perhaps, although we're not currently developing this library, so we've been holding off on new features until we can figure out where it's going.

nabijaczleweli:
So what's the status on this? Should I just use [a manual implementation](https://github.com/nabijaczleweli/tweetr/blob/457d6a27a0ba8293e9643443f08204e970cb7493/src/util.rs#L30-L53) and fuck off or...?

nabijaczleweli:
TBH I wanted this because `time::Duration` is reexported in `chrono::Duration`

alexcrichton:
@nabijaczleweli I'd recommend a local implementation for now, the libs team would probably prefer to overhaul this type entirely before piecemeal adding methods here and there for now.

nabijaczleweli:
OK then

PR #140, opened 2016-10-25
Not pull in win dependencies when not on windows

WiSaGaN:
This should avoid a bunch of unnecessary compiling (even download?) when not used on windows.


alexcrichton:
Thanks!

PR #142, opened 2016-12-22
Add support for Redox

jackpot51:
This adds support for Redox to the `time` crate

Due to the lack of appropriate time functions in `std`, and a desire to not be dependent on a `libc` - I have implemented `gmtime` and its inverse, `mktime` inline. Redox does not currently support timezones, so they will not be added to `time` until they are supported.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ Utilities for working with time-related functions in Rust.
 libc = "0.2.1"
 rustc-serialize = { version = "0.3", optional = true }
 
+[target.'cfg(target_os = "redox")'.dependencies]
+redox_syscall = "0.1"
+
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.0"
 kernel32-sys = "0.2.0"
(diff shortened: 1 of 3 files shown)


jackpot51:
@alexcrichton This will fix the compilation of, from what I can tell, about 400 directly dependent crates, and maybe thousands of indirectly dependent crates on Redox. The thing I expect to be questioned is the inlining of `gmdate`, but this allows for `time` to be used without dependencies on `libc` in Redox.

alexcrichton:
I talked about this briefly with @brson on IRC yesterday. I'm gonna hold off on merging this until we have a chance to discuss with @rust-lang/libs first in the next triage meeting.

My worry here is that adding a litany of os-specific dependencies has an impact on downstream users of this crate. For example if any crate runs `cargo vendor` or `cargo fetch` it'll pull *all* upstream dependencies regardless of the platform. I'd just want to make sure we take this step purposefully as we've discussed very little the impact of new OS support on the crates ecosystem, only for the standard library.

jackpot51:
@alexcrichton If all that were done were to add target_os = "redox" to the cfg(unix) line, would that be easier to merge?

jackpot51:
@alexcrichton is it possible to publish a new version to crates.io?

alexcrichton:
Sure, done.

PR #148, opened 2017-05-11
Workaround lack of tm_gmtoff on Solaris.

dhduvall:
The crate doesn't compile at all on Solaris due to a missing tm_gmtoff.  Until we get that implemented natively, we should have a workaround.  I'm not sure how to make it dynamic based on the presence or absence of the symbol in the struct, but this is at least a start.  It compiles and passes the testsuite.

@binarycrusader

### DIFF
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -209,6 +209,12 @@ mod inner {
     #[cfg(all(not(target_os = "macos"), not(target_os = "ios")))]
     pub use self::unix::*;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

nbaksalyar:
@alexcrichton could you please update `time` on crates.io with this patch? Cargo depends on this crate and currently some effort is required to make it work on Solaris/Illumos. Thank you!

alexcrichton:
Certainly!

PR #150, opened 2017-06-30
make time no_std compatible

th0br0:
There currently exists no `no_std` compatible time parser library. Let's change that!

This adds the `no_std` feature to the crate which selectively toggles (via far too many `#[cfg(...)]`) compilation against std or without.
Some tests were split or disabled if not compatible with `no_std`.


alexcrichton:
Thanks! Typically with no_std support that means the library is *always* tagged as `no_std` and then when the "use std" feature is enabled it'll pull in the standard library and add more impls, perhaps that route could be taken here? I think that'd help cut down on the amount of `#[cfg]` here perhaps.

alexcrichton:
I'm also wary for the "no_std" feature to start requiring nightly, would it be possiblef or all modes here to compile on stable?

th0br0:
Instead of having a `no_std` feature, there is now an `std` feature that is enabled by default.
`--no-default-features` works on stable, `to_string()` is only enabled on `nightly` given that `collections/alloc` is only available there. 
I'm guessing that those will stabilise at some point so that `ToString` is no longer in `std` on stable either?

src/lib.rs
```diff
@@ -29,12 +29,15 @@
...
 //! standard library.
-
 #![doc(html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
        html_favicon_url = "https://www.rust-lang.org/favicon.ico",
        html_root_url = "https://doc.rust-lang.org/time/")]
 #![allow(trivial_numeric_casts)]
+#![allow(unused_imports)]
 #![cfg_attr(test, deny(warnings))]
+#![cfg_attr(test, allow(deprecated))]
+#![cfg_attr(not(feature = "std"), no_std)]
```
alexcrichton: I'd actually recommend unconditionally having this attribute, and then all the import directives can import from `core` instead of `std` most of the time, hopefully cutting down on a lot of the `#[cfg]` traffic.

Cargo.toml
```diff
@@ -7,10 +7,16 @@ license = "MIT/Apache-2.0"
...
 Utilities for working with time-related functions in Rust.
 """
 
+[features]
+default = ["std"]
+std = []
+nightly = []
+
 [dependencies]
 libc = "0.2.1"
```
alexcrichton: Want to disable default features here so `libc` itself doesn't pull in the standard library?

src/lib.rs
th0br0: This conflicts with `rustc_serialize` which will not build if `no_std` is enabled (as it can't derive the required traits)

alexcrichton: Hm let me put this another way. This change should have `#![no_std]` unconditionally. Where the standard library is required there'd be a `#[cfg]` and you can have `extern crate std` and import directly from std.

alexcrichton: Oh this is baked into that derive mode... If that's the case then this can be conditional, but the library should have an explicit `extern crate core` and use imports from there rather than conditionally importing from core/std

src/duration.rs
```diff
@@ -401,11 +408,18 @@ impl fmt::Display for Duration {
...
 
 impl fmt::Display for OutOfRangeError {
+    #[cfg(feature = "std")]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", self.description())
     }
+
+    #[cfg(not(feature = "std"))]
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f,"Source duration value is out of range for the target type")
```
alexcrichton: Could this be deduplicated with the implementation above?

src/duration.rs
```diff
@@ -444,9 +458,18 @@ fn div_rem_64(this: i64, other: i64) -> (i64, i64) {
 #[cfg(test)]
 mod tests {
     use super::{Duration, MIN, MAX, OutOfRangeError};
-    use std::{i32, i64};
+
+    #[cfg(feature = "std")]
```
alexcrichton: Since testing always implies the standard library, could we just always run tests with the `std` feature?

src/lib.rs
```diff
@@ -45,10 +48,22 @@
...
-use std::ops::{Add, Sub};
+#[cfg(feature = "std")]
+mod std_impl;
+
+#[cfg(feature = "nightly")]
+#[macro_use]
+extern crate alloc;
+
+#[cfg(feature = "std")]
+extern crate core;
```
alexcrichton: Can this unconditionally always be here?

src/lib.rs
```diff
@@ -45,10 +48,22 @@
...
 #[cfg(all(windows, test))] extern crate advapi32;
 
-use std::cmp::Ordering;
-use std::error::Error;
-use std::fmt;
-use std::ops::{Add, Sub};
+#[cfg(feature = "std")]
+mod std_impl;
+
+#[cfg(feature = "nightly")]
```
alexcrichton: Could this feature be dropped for now? I'd prefer to not have the crate all of a sudden require a nightly compiler when certain features are enabled on the crate

src/lib.rs
```diff
@@ -144,142 +160,6 @@ impl Sub<Timespec> for Timespec {
...
-    Timespec::new(sec, nsec)
-}
-
-
-/**
- * Returns the current value of a high-resolution performance counter
- * in nanoseconds since an unspecified epoch.
- */
-pub fn precise_time_ns() -> u64 {
-    sys::get_precise_ns()
```
alexcrichton: Isn't this a valid api to have in a no-std situation?

src/duration.rs
th0br0: How'd we validate that the `no-std` implementation of, e.g. the formatting stuff, works as intended?The no-std doesn't know timezones, so needs some stuff handled separately.

src/lib.rs
th0br0: `rustc` complains about duplicate crate definition if it's there unconditionally.

src/lib.rs
th0br0: I understand your concern, but without this feature there's no way to do `strptime` or similar in no-std. The necessary structs are only available via alloc/nightly

src/lib.rs
th0br0: It depends on the underlying OS(or bare metal stack or whatever) though and thus can't be provided as part of the no-std library impl.

src/duration.rs
th0br0: We run into borrow-checker issues here as Error wants a `&str` but `Display/ToString` allocates a String. I've replaced the duplication with a const.

src/duration.rs
th0br0: Testing doesn't imply the standard library by default. I could drop the `cfg(feature)` for tests and always add std when testing.

alexcrichton:
Hm sorry I think I may be missing something. What's the use case of having the `time` crate be compatible with `no_std` if it doesn't actually expose any of the libc functionality?

th0br0:
A specific use case could be an embedded device that needs to parse or format a timestamp.
In addition to that, there's also the standardised/typed way of constructing a timestamp and accessing its different 'fields' (e.g. days, weeks, ... ) - in addition to being able to have a typed way of creating timestamps and accessing,

vks:
> Instead of having a no_std feature, there is now an std feature that is enabled by default.

Isn't that a breaking change for people who where using `time` with `no-default-features`?

th0br0:
The crate currently has no default features, so it wouldn't have made any sense for people to disable them. 
But yes, that would be a breaking change for whoever uses the crate with those settings and accesses methods for getting the current time / timezone stuff

alexcrichton:
@th0br0 but even with this PR there's still an unconditional dependency on libc? Doesn't that mean we should also take advantage of it?

th0br0:
Good catch - with 52b95ff libc is made an optional dependency.
`libc` is only used for the unix `sys::` methods

src/duration.rs
```diff
@@ -310,7 +356,10 @@ impl Add for Duration {
             nanos -= NANOS_PER_SEC;
             secs += 1;
         }
-        Duration { secs: secs, nanos: nanos }
+        Duration {
+            secs: secs,
+            nanos: nanos,
+        }
```
alexcrichton: It looks like there's a lot of formatting-related changes in this diff, can those all be removed into a separate PR?

src/duration.rs
alexcrichton: Ah sorry what I mean is that testing implies `extern crate test` which implies that you can have `extern crate std` here and then use imports like `ToString` from `std`. That'll avoid any need for `#[cfg]` here.

src/lib.rs
```diff
@@ -29,26 +29,36 @@
...
 #![allow(trivial_numeric_casts)]
+#![allow(unused_imports)]
 #![cfg_attr(test, deny(warnings))]
+#![cfg_attr(test, allow(deprecated))]
+#![cfg_attr(not(feature = "std"), no_std)]
 
 #[cfg(target_os = "redox")] extern crate syscall;
-#[cfg(unix)] extern crate libc;
+#[cfg(all(unix, feature = "std"))] extern crate libc;
 #[cfg(windows)] extern crate kernel32;
```
alexcrichton: I think these run the same problem with `libc`, right? This effectively means that `no_std` doesn't work on Windows?

src/sys.rs
```diff
@@ -2,14 +2,18 @@
...
+#[cfg(any(target_os = "redox", not(feature = "std")))]
 mod inner {
-    use std::fmt;
-    use std::cmp::Ordering;
-    use std::ops::{Add, Sub};
+    use core::fmt;
+    use core::cmp::Ordering;
+    use core::ops::{Add, Sub};
+
+    #[cfg(feature = "std")]
```
alexcrichton: This entire module is gated, right? Does each statement here need a `#[cfg]`? Is there a way we can cut down on the `#[cfg]` traffic?

PR #165, opened 2018-05-10
Add Support for wasm32-unknown-unknown

CryZe:
Most things just panic like std does. The time_to_tm and tm_to_time implementations are shared with the Redox implementations.


CryZe:
Thanks for merging :)
Can we get a release for this?

alexcrichton:
Published!

PR #168, opened 2018-09-06
Add #[derive]d Hash impl to Duration

nabijaczleweli:
This counts as a "bugfix", right?


giacomocariello:
It would be great if this could fit in the next release cycle.

PR #172, opened 2018-12-28
Fix issue with SGX port, bump crate version

jethrogb:


jethrogb:
I don't think the CI failure is caused by this PR

PR #181, opened 2019-09-23
Remove the patch revision specification for libc

AdminXVII:
Making the dependency on "libc" specific to the patch revision removes the ability to get new updates. For example, the 0.2.62 version is needed for Redox to work, but this crate does not allow it.

Remove the patch version and also bump the patch version of the crate for release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "time"
-version = "0.1.42"
+version = "0.1.43"
 authors = ["The Rust Project Developers"]
 license = "MIT/Apache-2.0"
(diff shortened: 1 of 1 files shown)


AdminXVII:
Hi, is there a reason why this was closed? Will it be fixed in another PR? Because that's blocking many things on Redox, so it would be nice to fix it.

jhpratt:
@AdminXVII I'm currently in the process of rewriting time to be based on `std::time`. v0.1 is essentially unmaintained at this point. Hopefully there will be something relatively soon that can fix your issues!

AdminXVII:
@jhpratt Ok, thanks! Glad to hear this.

PR #185, opened 2019-10-21
Fix typos in README

clee:
Also enabled TOML syntax highlighting for TOML snippets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,16 +12,16 @@
 
 ### `#![no_std]`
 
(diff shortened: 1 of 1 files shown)


jhpratt:
I was going to say that "useable" is preferred to "usable", but Merriam-Webster has the former as a less common variant. For whatever reason, I didn't realize that TOML was supported as a "language" in GFM.

Thanks! Merging.

PR #187, opened 2019-11-16
Localization: Add French

homersimpsons:


jhpratt:
Are the month and weekday names always capitalized in French? From quickly searching, that doesn't appear to be the case. For months, on [what appears to be the official website](http://www.academie-francaise.fr/majuscules-aux-noms-de-jours-et-de-mois) (using Google Translate; I don't speak French):

> Since they are common nouns, they must not, except at the beginning of the sentence, be written with a capital letter and they take, one and the other, the mark of the plural.

Also, what level experience do you have with French (personally)?

homersimpsons:
@jhpratt You are right, months and weekdays are not capitalized in french. I did copy paste the english translation and though that everything should be capitalize (I didn't see that spanish ones weren't).

I am French so i'm a native speaker :)

(I will fix the PR tomorrow)

jhpratt:
Awesome, thanks! I'll merge it in once that's updated.

homersimpsons:
I did just push an update.

I ordered the enum alphabetically and use the same translations as C `strftime` ([found here](https://sourceware.org/git/?p=glibc.git;a=blob;f=localedata/locales/fr_FR;h=a18c514f1921fed0049d3b769c95c9e0f864fb2f;hb=HEAD)) as the doc states:

> Time's formatting behavior is based on strftime in C

jhpratt:
Complete nitpick, but can the periods/full stops be removed from the abbreviations?

Side note: I wish GitHub gave me some way to edit the code directly. I _could_ merge the PR with edits locally, but then this wouldn't be indicated as merged.

homersimpsons:
I just removed those periods. Those are used in french to indacate that it's an avbbreviation in that context.

jhpratt:
It's the same situation in English, generally speaking. I just think that it's trivial to add the periods if desired, but it's not as easy to remove them.

Thanks for the changes! Merging.

homersimpsons:
But for some short month the abbreviation is the same as the none abbreviated form so it might be harder to add correctly the period if needed :)

jhpratt:
Hmm, you're right. I'll think about this more for sure, as that's also the case for English. Perhaps it could be another modifier on the specifier.

jhpratt:
@homersimpsons After looking into glibc some, it looks like their localization files are quite accurate. As such, I'll actually be pulling in data for plenty of other languages as well, leaving punctuation as-is.

This will be done via a build script, as otherwise it would be an unwieldy file to maintain.

Thanks again for the reference.

PR #188, opened 2019-11-19
Put 'non_exhaustive' attribute behind feature flag

jpopesculian:
This is a problem when compiling for stable versions of rust

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,10 @@ license = "MIT OR Apache-2.0"
 # TODO Add GitHub Actions badge once rust-lang/crates.io#1838 is merged.
 
 [features]
-default = ["std", "deprecated"]
+default = ["std", "deprecated", "unstable"]
 deprecated = ["log"]
 std = []
+unstable = []
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


jhpratt:
This behavior is intentional. Minimum supported Rust version is currently beta (1.40), as indicated by the README. This change was done in b9b1b74, as otherwise a hidden field or variant was necessary to simulate the attribute.

Given that this won't be released before 1.40, I'm leaning towards closing this, unless you can think of a reason not to? Merging it would break back-compatibility once 1.40 stabilizes.

jpopesculian:
Ah I see, I didn't notice the 1.40 badge. Its fairly specific to my use case that I can't use unstable features, and I think it makes sense what you're saying.

jhpratt:
Yeah, the version isn't released yet. You could pin the commit in the Cargo.toml to before the one referenced above, which will definitely compile on stable. If you want the more recent and upcoming changes, I imagine forking and applying this PR wouldn't too too difficult. You'd have to remember to switch back after 1.40 is released, though.

Thanks for the understanding!

PR #195, opened 2019-12-13
fix redox build by added additional target_os check for unixes

njskalski:
one way to partially fix https://gitlab.redox-os.org/redox-os/redox/issues/1279

### DIFF
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -347,7 +347,7 @@ mod inner {
     }
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "redox")))]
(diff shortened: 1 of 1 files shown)


jhpratt:
v0.2 of time will likely be released in under a week. I believe Redox has its own fork on its GitLab that has Redox-specific patches; is that not correct? Hopefully it won't be necessary soon, though.

jackpot51:
Chrono at the latest version is still using time 0.1. Many other crates are still using at, and will for the foreseeable future. Please consider merging this simple change to the v0.1 branch and making a new 0.1 release.

PR #206, opened 2020-01-27
Symlink licenses into subcrates

cuviper:
These licenses usually require that they are distributed along with the
sources they apply to, but they are currently missing from the crates.io
packages for `time-macros` and `time-macros-impl`. Simple symlinks here
will solve this if you're publishing from unix-like platforms, but
they'll need complete copies for this to work from Windows.

### DIFF
--- a/time-macros-impl/LICENSE-Apache
+++ b/time-macros-impl/LICENSE-Apache
@@ -1,0 +1,1 @@
+../LICENSE-Apache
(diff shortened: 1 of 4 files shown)


jhpratt:
I work from Ubuntu, so symlinks work for me. Nice catch, though, didn't think of downstream requirements.

PR #209, opened 2020-01-28
[v0.1] Fix warnings and test dates

cuviper:
While time-0.1 is not actively maintained anymore, these warnings might cause issues if there's ever a need to issue a patch release, so I think they're worth fixing. `SOME_FUTURE_DATE` also caused an amusing failure since we're now in that future, so I've bumped that date out to [Y2038](https://en.wikipedia.org/wiki/Year_2038_problem). Everything is still compatible with Rust 1.21.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ sudo: false
 matrix:
   include:
     - rust: 1.21.0
+      before_script:
+        - cargo generate-lockfile
+        - cargo update -p cfg-if --precise 0.1.9
     - rust: stable
     - os: osx
     - rust: beta
--- a/src/display.rs
+++ b/src/display.rs
@@ -11,9 +11,9 @@ impl<'a> fmt::Display for TmFmt<'a> {
                     if ch == '%' {
                         // we've already validated that % always precedes
                         // another char
-                        try!(parse_type(fmt, chars.next().unwrap(), self.tm));
+                        parse_type(fmt, chars.next().unwrap(), self.tm)?;
                     } else {
-                        try!(fmt.write_char(ch));
+                        fmt.write_char(ch)?;
                     }
                 }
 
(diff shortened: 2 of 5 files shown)


cuviper:
I notice that master has switched from travis to github workflows. Should I change that here too?

jhpratt:
Was MSRV 1.21 before? If so, I'll definitely merge this in after ensuring everything works as expected.

And feel free to switch to GitHub Actions! I find it easier to work with, plus it's on the same site.

cuviper:
> Was MSRV 1.21 before?

I don't see MSRV explicitly stated anywhere, but that's the version tested since b45c8a51bb9d285f96e14cfad2cf5fde9f279d44.

jhpratt:
1.21 works for me then, once tests pass. I'm assuming you just pulled from the current master (which is fine).

cuviper:
Yeah, I started with master's workflow, and pruned it back to the basics. Looks like I've got more warnings on Mac and Windows to address now -- I had only tried Linux myself. :)

cuviper:
Windows tests are failing in `acquire_privileges` to set the time zone -- maybe the environment is different in GiHub Actions than it was in AppVeyor? How would you feel about skipping windows tests? (would still `cargo check`, at least)

jhpratt:
Let's skip it for now. I can take a look into things later, possibly filing an issue on [actions/virtual-environments](https://github.com/actions/virtual-environments).

cuviper:
```
assert_eq!(tkp.tkp.Privileges.len(), 0);
```

I think this has been broken since winapi 0.3.0 due to [this change](https://github.com/retep998/winapi-rs/commit/40147475240fc9095e773598b6f30f3a4dda6504#diff-18c1aee6a909aea6ee58a714b406b114L1308-R2408) from array length 0 to `ANYSIZE_ARRAY`, which is 1. I wonder why subsequent CI never caught this? But I have a potential fix to try...

cuviper:
:heavy_check_mark: :tada:

jhpratt:
Awesome! I'll review and merge this in later.

These changes don't affect anything at runtime, so a new release isn't necessary, right?

cuviper:
> These changes don't affect anything at runtime, so a new release isn't necessary, right?

Correct. The main build just had some warnings, and even those will be masked as a dependency because cargo uses `--cap-lints allow`. Should be low risk if you do publish though.

As for MSRV, it actually can compile all the way back to Rust 1.13, though you might choose not to make that guarantee. My change to use `?` here won't work on 1.12, but also the crate was already using attributes on non-item statements which needs 1.13 anyway.

PR #217, opened 2020-02-14
Method to get local UTC offset at a given datetime

jhpratt:
This implementation should work on Windows, Mac, Linux, and Solaris. As Redox is *nix, it should work there as well.

`OffsetDateTime::now_local()` returns a value with the local offset.

This necessarily requires syscalls, and as such `#![forbid(unsafe_code)]` has been changed to `#![deny(unsafe_code)]`. It must be explicitly `#[allow]`ed in every location it is used, along with a general message describing why the usage is safe.

### DIFF
--- a/src/offset_date_time.rs
+++ b/src/offset_date_time.rs
@@ -42,9 +42,6 @@ pub struct OffsetDateTime {
 impl OffsetDateTime {
     /// Create a new `OffsetDateTime` with the current date and time.
     ///
-    /// This currently returns an offset of UTC, though this behavior will
-    /// change once a way to obtain the local offset is implemented.
-    ///
     /// ```rust
     /// # use time::{OffsetDateTime, offset};
     /// assert!(OffsetDateTime::now().year() >= 2019);
@@ -926,7 +923,8 @@ impl From<SystemTime> for OffsetDateTime {
                 .expect("overflow converting `std::time::Duration` to `time::Duration`"),
         };
 
-        Self::unix_epoch() + duration
+        let t = Self::unix_epoch() + duration;
+        t.to_offset(UtcOffset::local_offset_at(t))
     }
 }
 
@@ -957,7 +955,10 @@ mod test {
     #[cfg(feature = "std")]
     fn now() {
         assert!(OffsetDateTime::now().year() >= 2019);
-        assert_eq!(OffsetDateTime::now().offset(), offset!(UTC));
+        assert_eq!(
+            OffsetDateTime::now().offset(),
+            UtcOffset::current_local_offset()
+        );
     }
 
     #[test]
--- a/src/offset_date_time.rs
+++ b/src/offset_date_time.rs
@@ -40,12 +40,12 @@ pub struct OffsetDateTime {
 }
 
 impl OffsetDateTime {
-    /// Create a new `OffsetDateTime` with the current date and time in the
-    /// local offset.
+    /// Create a new `OffsetDateTime` with the current date and time in UTC.
     ///
     /// ```rust
     /// # use time::{OffsetDateTime, offset};
     /// assert!(OffsetDateTime::now().year() >= 2019);
+    /// assert_eq!(OffsetDateTime::now().offset(), offset!(UTC));
     /// ```
     #[inline(always)]
     #[cfg(feature = "std")]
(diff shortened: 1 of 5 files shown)


jhpratt:
NB: This is the only thing holding up a 0.2.7 release. What I'm looking for is one or two more people to look everything over, and have at least one person verify the implementation works as expected on Windows.

src/offset_date_time.rs
```diff
@@ -40,15 +40,12 @@ pub struct OffsetDateTime {
 }
 
 impl OffsetDateTime {
-    /// Create a new `OffsetDateTime` with the current date and time.
-    ///
-    /// This currently returns an offset of UTC, though this behavior will
-    /// change once a way to obtain the local offset is implemented.
```
aloucks: Would you consider having `now()` continue to return the time in UTC and instead implement a `now_local()`?

jhpratt: My thinking behind having `now()` return local by default is that if someone does `OffsetDateTime::now().hour()`, they're more than likely expecting _their_ hour, not UTC.

Equality is implemented in terms of the actual time (UTC), so the two methods will actually return equal values (ignoring the difference in time between calling them).

aloucks: All existing usage of `now()` will be silently broken with this change. In addition, the source of most bugs (in my experience) of code involving timestamps is often related to trying to juggle timezones or converting back and forth. I always recommend storing and using timestamps in UTC and only converting to local time in the final step before presenting to a user. The original implementation (where `now()` was UTC) encouraged this behavior. I fear that making `now()` return local time is a step backwards.

jhpratt: The internal usage _is_ stored as UTC. It's only converted to the stored offset as necessary.

How will existing uses of `now()` be broken, given that all equality, comparisons, hashing, etc. are implemented in terms of the timestamp? The _only_ thing changing is the offset.

aloucks: Broken in the sense that prior to the change the offset was UTC and now it's local timezone. I just mean that it's a breaking change. 

> The internal usage is stored as UTC

I understand that. I'm saying that `now() -> UTC` encourages others to also use UTC.

jhpratt: It was and is clearly documented that the behavior would change, so anyone relying on that shouldn't be. If they need UTC, they should already be doing `.to_offset(offset!(UTC))`.

I will consider this, however.

Given that I still need eyes to look everything over for safety and someone to ensure everything works on Windows (and have been asking for around a week), I likely have a bit before a decision must be made.

jhpratt: Updated to have `now()` return UTC and `now_local()` return local. This is in accordance with the "vision" page in the wiki.

Thanks!

kevinpoitra:
Thanks for adding this! I'm looking to upgrade one of the MySQL libraries to use `time` 0.2, and I need to be able to get the local UTC offset for some of its SQL `DATETIME` -> `OffsetDateTime` type conversion stuff (it could probably just use UTC, but the implementation already uses the local offset via `time` 0.1's `at(Timespec::new(0, 0)).tm_utcoff`, so I figure it's best to keep it the way it is).

I'm using Windows 10, and everything seems to work as expected. To test, I created a fresh binary with only `time` as a dependency, like so:
```toml
time = { git = "https://github.com/time-rs/time", branch = "local-offset", default-features = false, features = ["std"] }
```

I then wrote a few quick `println`s based on the functions added within your commits:

```Rust
use time::{OffsetDateTime, UtcOffset};

fn main() {
    let local_utc_offset = UtcOffset::current_local_offset();
    println!("Local UTC Offset from UtcOffset::current_local_offset(): {}", local_utc_offset);

    println!("Local UTC Offset from OffsetDateTime::now().offset(): {}", OffsetDateTime::now().offset());
    println!("Local UTC Offset from OffsetDateTime::now_local().offset(): {}", OffsetDateTime::now_local().offset());

    println!("Local UTC Offset from UTC time: {}", UtcOffset::local_offset_at(OffsetDateTime::now()));
    println!("Local UTC Offset from Local time: {}", UtcOffset::local_offset_at(OffsetDateTime::now_local()));

    println!();

    println!("Now, UTC: {}", OffsetDateTime::now());
    println!("Now, Local: {}", OffsetDateTime::now_local());

    let unix_epoch = OffsetDateTime::unix_epoch();
    println!("Unix Epoch, UTC: {}", unix_epoch);
    println!("Unix Epoch, UTC {}: {}", local_utc_offset, unix_epoch.to_offset(local_utc_offset));
}
```

The machine is located within the UTC -6 timezone, and the results match up with what you'd expect them to be:
```
Local UTC Offset from UtcOffset::current_local_offset(): -6
Local UTC Offset from OffsetDateTime::now().offset(): +0
Local UTC Offset from OffsetDateTime::now_local().offset(): -6
Local UTC Offset from UTC time: -6
Local UTC Offset from Local time: -6

Now, UTC: 2020-02-22 11:47:23.228243200 +0
Now, Local: 2020-02-22 5:47:23.228243200 -6
Unix Epoch, UTC: 1970-01-01 0:00 +0
Unix Epoch, UTC -6: 1969-12-31 18:00 -6
```

If there's anything else you'd like me to test out, feel free to ping me!

jhpratt:
Thanks a ton @kevinpoitra! Looks like everything works as expected on all OS's then. I'll merge this in later today.

Just one note: Keep in mind that `at(Timespec::new(0, 0)).tm_utcoff` would have to be translated to `UtcOffset::local_offset_at(OffsetDateTime::unix_epoch())`, as otherwise you'd be getting the _current_ offset, which isn't the same during the summer for a number of locations.

PR #224, opened 2020-02-24
Support for local offsets with cargo_web

### LINKED ISSUE #223: Doesn't build anymore for webassembly target
I'm using the time package as a transitive dependency. The latest build seems incompatible with webassembly. Building my project with "cargo web build" yields the following error message:

       Compiling time v0.2.7
    error[E0308]: mismatched types
       --> /home/hc/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.2.7/src/utc_offset.rs:294:53
        |
    294 | fn try_local_offset_at(datetime: OffsetDateTime) -> Option<UtcOffset> {
        |    -------------------                              ^^^^^^^^^^^^^^^^^ expected enum `std::option::Option`, found `()`
        |    |
        |    implicitly returns `()` as its body has no tail or `return` expression
        |
        = note:   expected enum `std::option::Option<utc_offset::UtcOffset>`
                found unit type `()`

hce:
Here is my attempt to fix #223

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ readme = "README.md"
 license = "MIT OR Apache-2.0"
 # TODO(upstream) Add GitHub Actions badge once rust-lang/crates.io#1838 is merged.
 description = "Date and time library. Fully interoperable with the standard library. Mostly compatible with #![no_std]."
+build = "build.rs"
 
 [package.metadata.docs.rs]
 all-features = true
@@ -26,6 +27,7 @@ std = ["libc", "winapi"]
 __doc = []
 
 [dependencies]
+cfg-if = "0.1"
 rand = { version = "0.7", optional = true, default-features = false }
 rustversion = "1"
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }
@@ -41,4 +43,4 @@ libc = { version = "0.2", optional = true }
 winapi = { version = "0.3", features = ["minwinbase", "minwindef", "timezoneapi"], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-stdweb = "*"
+stdweb = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ all-features = true
 default = ["deprecated", "std"]
 deprecated = []
 panicking-api = []
-std = ["libc", "winapi"]
+std = ["libc", "winapi", "stdweb"]
 
 # Internal usage. This is used when building for docs.rs and time-rs.github.io.
 # This feature should never be used by external users. It will likely be
@@ -43,4 +43,4 @@ libc = { version = "0.2", optional = true }
 winapi = { version = "0.3", features = ["minwinbase", "minwindef", "timezoneapi"], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-stdweb = "0.4"
+stdweb = { version = "0.4", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,3 +39,6 @@ libc = { version = "0.2", optional = true }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["minwinbase", "minwindef", "timezoneapi"], optional = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+stdweb = "*"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -39,3 +39,6 @@ libc = { version = "0.2", optional = true }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["minwinbase", "minwindef", "timezoneapi"], optional = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+stdweb = "*"
```
jhpratt: This should use a standard version number, not a glob import. Otherwise, breakage could happen at any time with an upstream change. Using `default_features = false` would also be a good idea if possible.

src/utc_offset.rs
```diff
@@ -8,6 +8,9 @@ use crate::{
 };
 use core::fmt::{self, Display};
 
+#[cfg(target_arch = "wasm32")]
+use stdweb::js;
```
jhpratt: _Major_ point here: `stdweb` does not work on any wasm target. wasmtime, for example, would have no idea what to do.

src/utc_offset.rs
```diff
@@ -448,8 +452,28 @@ fn try_local_offset_at(datetime: OffsetDateTime) -> Option<UtcOffset> {
...
+    {
+        let timestamp_utc: i64 = datetime.timestamp();
+        let t1 = (timestamp_utc & 0xFFFFFFFF) as i32;
+        let t2 = (timestamp_utc >> 32) as i32;
+        let timezone_offset = js! {
+            var ts_utc = @{t2} << 32 + @{t1};
+            var d = new Date(ts_utc);
+            return d.getTimezoneOffset() * -60;
+        };
+        if let Ok(seconds) = stdweb::unstable::TryInto::try_into(timezone_offset) {
```
jhpratt: Is this `TryInto` any different than `std::conert::TryInto`?

We could also likely use `timezone_offset.try_into().ok()?.map(UtcOffset::seconds)` here.

src/utc_offset.rs
```diff
@@ -448,8 +452,28 @@ fn try_local_offset_at(datetime: OffsetDateTime) -> Option<UtcOffset> {
...
+            return Some(UtcOffset::seconds(seconds));
+        } else {
+            return None;
+        }
+    }
+
+    #[allow(unreachable_code)]
+    None
```
jhpratt: I'd prefer to avoid having this just floating around, especially with allowing that lint. I think it would be cleaner to aggressively gate it.

`cfg_if` will likely be helpful here, especially as the number of targets increases.

src/utc_offset.rs
```diff
@@ -448,8 +452,28 @@ fn try_local_offset_at(datetime: OffsetDateTime) -> Option<UtcOffset> {
...
+
+    #[cfg(target_arch = "wasm32")]
+    {
+        let timestamp_utc: i64 = datetime.timestamp();
+        let t1 = (timestamp_utc & 0xFFFFFFFF) as i32;
+        let t2 = (timestamp_utc >> 32) as i32;
+        let timezone_offset = js! {
+            var ts_utc = @{t2} << 32 + @{t1};
+            var d = new Date(ts_utc);
+            return d.getTimezoneOffset() * -60;
```
jhpratt: This could realistically be combined into a single statement. I imagine `stdweb` doesn't do any sort of optimization on the output.

src/utc_offset.rs
hce: I agree it's a problem. The trouble is, wasm is a very generic standard that defines no default interface to interact with the outside world. I will check if it is possible to narrow this down to a javascript environment.

hce:
I tried to implement all the changes you requested.

It seems it is not possible to dynamically add a dependency from build.rs, so unfortunately, I have to include the stdweb package as a dependency whenever the target is wasm32, but; I think the rust linker optimizes unused libraries out of the final result, so we should be fine.

jhpratt:
It's definitely not possible enable/disable dependencies from a build script. I'll take a look to see what other crates have done; hopefully there's some way to avoid it altogether.

Did you check to see if `default_features = false` works here? If not, I can do so when I check it out locally.

hce:
No, I didn't check the default_features flag yet. The problem is that cargo-web only sets an environment variable to allow to detect its environment; I don't see how default_features could help here...

jhpratt:
Yeah, looks like that's the case. And `default_features` just reduces the number of crates needing to be built, it has nothing to do needing a build script.

I'll checkout the branch and take a final look. Everything looks good at first, though.

jhpratt:
All should be set! Thanks for this — I'm actually looking into adding support for WASI now.

PR #225, opened 2020-02-25
JS' new Date() expects milliseconds + fix operator precedence

hce:
I'm sorry, I checked the function some more and turns out I made two errors: The operator precedence is not what I assumed in JS plus "new Date(...)" expects EPOCH in milliseconds, not seconds.


jhpratt:
You're right; I'd completely forgotten that JavaScript takes values in milliseconds. Guess that's what happens when you haven't touched it in a few months.

Thanks!

Issue #228, opened 2020-03-01
Parsing year problem


SLMT:
I got a problem when I parsed a string with format `%Y%m%d`. Here is a test case:

```rust
use time::Date;

fn main() {
  let date = Date::parse("20200703", "%Y%m%d");
  dbg!(date);
}
```

This prints:

```
[src/main.rs:5] date = Err(
    InvalidYear,
)
```

But, it works fine in the following case:

```rust
use time::Date;

fn main() {
  let date = Date::parse("2020 0703", "%Y %m%d");
  dbg!(date);
}
```

```
[src/main.rs:5] date = Ok(
    Date {
        year: 2020,
        ordinal: 185,
    },
)
```

It also looks like it can distinguish the month and the day when they come together, but it cannot do the same thing for the year. So, I guess this might be a bug?

SLMT:
I look into the code of parsing years. It looks like it assumes the string of a year might have at most 6 digits so that it can handle a year greater than 9999. However, I think may be better to assume the string is 4 digits and accept a string that has more than 4 digits only when it begins with a `+`.

How do you think about this?

jhpratt:
I've actually been aware this would likely happen, but never bothered to test it. You're 100% correct that the parser assumes can be up to six digits, as the valid years are ±100,000.

Given that the formatter inserts a `+` (per the ISO spec) when the year is more than four digits, I think parsing it the same way would make sense. It should be a relatively minor change, just adding a conditional to change the max from 6 to 4 if a `+` is parsed. I can do this at some point, likely after Tuesday (midterms for college).

SLMT:
How about I do it later and submit a pull request?

jhpratt:
If you'd like, sure! Just be sure to add a test case.

SLMT:
I just submitted a pull request. Please check if it is good for you.

PR #229, opened 2020-03-01
Restrict parsed years to 4 digits unless sign is present

SLMT:
This pull request fixes the problem identified in #228 .

I made the parser accept a year string with more than 4 digits only when the string has a sign (+/-), so that it can parse a string where a 4-digits year and other numbers are put together without spacing. For example: `Date::parse("20200201", "%Y%m%d")`.

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -2176,10 +2176,7 @@ mod test {
             Date::parse("2019-W01-3", "%G-W%V-%u"),
             Ok(date!(2019-W01-3))
         );
-        assert_eq!(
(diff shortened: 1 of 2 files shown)


jhpratt:
Quick look is good! I'll take a look once I'm on my laptop and merge this in later.

Thanks!

jhpratt:
LGTM and all checks pass. Merging.

SLMT:
Thanks! BTW, this is great project. It helps me a lot.

PR #231, opened 2020-03-10
Hack around a getting precise time in Windows

da-x:
This takes the approach from https://github.com/rust-lang/rust/issues/67266 by @dbergman. This may be necessary for the v0.1 branch of `time` because the widely-used `chrono` crate still depends on it. When writing logs on Windows, crates such as `simplelog` which depend on `chrono` would benefit from a better timestamp. @quodlibetor @Drakulix

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ documentation = "https://doc.rust-lang.org/time"
 description = """
 Utilities for working with time-related functions in Rust.
 """
+build = "build.rs"
 
 [dependencies]
 libc = "0.2.1"
@@ -20,6 +21,9 @@ redox_syscall = "0.1"
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.0", features = ["std", "minwinbase", "minwindef", "ntdef", "profileapi", "sysinfoapi", "timezoneapi"] }
+
+[build-dependencies]
+rustc_version = "0.2"
 
 [dev-dependencies]
 log = "0.4"
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,8 @@
+extern crate rustc_version;
+use rustc_version::{version, Version};
+
+fn main() {
+    if version().unwrap() >= Version::parse("1.31.0").unwrap() {
+        println!("cargo:rustc-cfg=feature=\"has_const_fn\"");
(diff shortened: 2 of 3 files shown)


jhpratt:
A few things:

- CI is failing due to the use of a static atomic
- I don't have knowledge of Windows APIs. Please comment the code so I can follow what it's doing and why.
- This code does not appear to be compatible with Windows < 8, per the rust-lang issue.
- What benefit does this provide, exactly? Time v0.1 is no longer supported, so only major issues are being fixed.

da-x:
* Seems to be failing due to Rust 1.24 compatibility. I'll look into resolving it.
* Sure, I'll add the comments.
* It should be compatible with older versions of Windows, as it implements a fallback to the older function if the newer function is not exported from kernel32. The pending PR I've opened for `std` does something similar but in a cleaner approaching using existing `std` machinery.
* The major benefit is the accuracy of a timestamp going down significantly from 10ms to the 1usec area. This is useful for applications that write log files.

jhpratt:
I'm still not quite sure this is worthy of another release. As I mentioned, 0.1 isn't supported any more. The current limitation of ~10ms is how it's been for a very long time. If this is released in a 0.1.43, there are a number of other things that I have previously closed that people would want to revisit.

Is there anywhere this has caused major issues that can't be worked around?

da-x:
I guess that a possible workaround is to bake this functionality straight into the `chrono` crate, and have it not depend on old `time`, maybe only for Windows. Maybe @quodlibetor can comment?

jhpratt:
Pinging @da-x and @quodlibetor for an update. This PR will be closed without anything new — I have no desire to maintain 0.1, as it hasn't been touched significantly in years.

jhpratt:
As I decided in #240, I'll be creating one final release for 0.1 at some point in the future. As such, I'll review this PR along with others when I get the chance.

pfmooney:
>  It should be compatible with older versions of Windows, as it implements a fallback to the older function if the newer function is not exported from kernel32. The pending PR I've opened for `std` does something similar but in a cleaner approaching using existing `std` machinery.

It would be nice to have a test plan for these older versions of Windows, both with stable rust and 1.21 (if possible).

PR #234, opened 2020-03-13
Force cfg-if 0.1.10

### LINKED ISSUE #233: time 0.2.8 fails to compile when cfg-if is pinned to 0.1.9
I'm getting the following error when trying to compile using time 0.2.8. time 0.2.7 works just fine.

```
error: expected an item keyword
   --> /Users/bryan/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.2.8/src/utc_offset.rs:366:13
    |
366 |             let tm = timestamp_to_tm(datetime.timestamp())?;
    |             ^^^

error: could not compile `time`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

So far, this is as minimal of a reproduction as I can get. The following `Cargo.toml` with an empty lib.rs file.

```toml
[package]
name = "example"
version = "0.1.0"
edition = "2018"

[dependencies]
#time = "=0.2.7"
time = "=0.2.8"
phonenumber = "0.2"
```

bryanburgers:
When using cfg-if version `0.1.9`, compiling fails with the following
error message.

    error: expected an item keyword
       -->
    .../time-0.2.8/src/utc_offset.rs:366:13
        |
    366 |             let tm = timestamp_to_tm(datetime.timestamp())?;

Forcing 0.1.10 fixes this.

Fixes #233

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ std = ["libc", "winapi", "stdweb", "standback/std"]
 __doc = []
 
 [dependencies]
-cfg-if = "0.1"
+cfg-if = "0.1.10"
 rand = { version = "0.7", optional = true, default-features = false }
 rustversion = "1"
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }


jhpratt:
As I said in the issue, I'm going to do this as part of a larger effort to ensure it doesn't break unexpectedly with other (non-current) versions of direct dependencies.

bryanburgers:
Yeah, no problem. I was already about to submit this when I saw your comment, so I submitted it anyway. No hard feelings if you don't use it.

For now, adding

```toml
[patch.crates-io]
time = { git = "https://github.com/bryanburgers/time", branch = "force-cfg-if-0.1.10" }
```

gets my project to compile successfully. Looking at the Cargo.lock file, it looks like Cargo is downgrading lexical-core from 0.6.7 to 0.6.2 to satisfy all of the dependencies 🤷‍♂.

jhpratt:
Looks like this is the only dependency that causes issues.

PR #240, opened 2020-04-03
Remove special casing for Redox, as it is now supported by Unix target family

jackpot51:
This is an alternative to #195 which utilizes libc on Redox. `time` 0.1 is still used all over the place so it would be great if it could compile on Redox. This change fixes issues with matching both `cfg(unix)` and `cfg(target_os = "redox")` by removing the unnecessary additional code for Redox.


jhpratt:
Time 0.1 is no longer supported, but I will investigate one final round of changes for a 0.1.43 release. This is _only_ because 0.1 has known issues and evidently does not currently build on some systems. Consider this the exception, rather than the rule.

I would like to firmly establish the MSRV, which I know is at least 1.21. There are likely a number of changes that could be made — closed PRs is likely a good starting place. Between COVID-19, college, family, and other commitments, I am very busy, so this certainly won't happen overnight.

Feel free to join Matrix if there is anything specific you'd like to discuss.

pfmooney:
Hey @jackpot51,
@jclulow  and I are picking up maintenance of the 0.1 branch.  Are you at all concerned about Redox users compiling for older versions of rust/libc where the necessary bits aren't yet defined, or is the community keeping pretty up to date on that?  If it's the latter, we can get this merged and rolled into the next patch release.

jackpot51:
No, there is no concern about that

jackpot51:
Thanks for merging!

PR #246, opened 2020-04-18
`#![no_std]`, embedded support

### LINKED ISSUE #243: Can't build with no_std because of cfg-if dependency
```
 └─  cargo c
    Updating crates.io index
       F       Fetch [================>                                      ]  31.39%                                                                                                        Downloaded time-macros v0.1.0
  Downloaded time-macros-impl v0.1.0
  Downloaded standback v0.2.2
  Downloaded proc-macro-hack v0.5.15
  Downloaded time v0.2.9
  Downloaded 5 crates (123.9 KB) in 0.91s
   Compiling syn v1.0.17
   Compiling rustversion v1.0.2
   Compiling standback v0.2.2
   Compiling proc-macro-hack v0.5.15
    Checking cfg-if v0.1.10
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv7em-none-eabihf` target may not be installed

For more information about this error, try `rustc --explain E0463`.
error: could not compile `standback`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

PTaylor-FluenTech:
Fixes #243 
Requires jhpratt/standback#6

This should allow `time` to be used with no_std targets **assuming** the author of standback accepts my PR.

The second commit isn't required, but (in my opinion) is a cleaner way of checking the `std` feature. It might be completely wrong.

### DIFF
--- a/time-macros/src/lib.rs
+++ b/time-macros/src/lib.rs
@@ -1,3 +1,5 @@
+#![no_std]
+
 use proc_macro_hack::proc_macro_hack;
 
 #[proc_macro_hack]
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,9 +8,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        rust: [1.34.0, 1.36.0, stable, nightly]
+        rust: [1.34.0, 1.36.0, stable]
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        target: [thumbv7em-none-eabihf]
 
     steps:
       - name: Checkout sources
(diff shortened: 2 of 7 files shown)


PTaylor-FluenTech:
Waiting for resolution of jhpratt/standback#6 and version bump.

time-macros/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![no_std]
```
PTaylor-FluenTech, line 1: I believe this is ok even when this crate is used by other projects that are not no_std.

jhpratt:
Standback is my crate, FYI. Consider that done after a quick review later.

The second commit is correct in functionality, but I felt it was cleaner to emit two separate flags. I'm open to being convinced, though!

As with standback, I'll handle the CI side of things to avoid unnecessary duplication.

time-macros/src/lib.rs
jhpratt: That's correct. I'll check the semantics on whether this is necessary, as it's only present at build time.

PTaylor-FluenTech:
> Standback is my crate, FYI.

Boy do I feel silly 😳.

>The second commit is correct in functionality, but I felt it was cleaner to emit two separate flags. I'm open to being convinced, though!

As I've mentioned, I'm pretty new. I think I've seen it as in my commit more, but don't have strong feelings about it. I think for me, it doesn't _hide_ the `no_std` flag. When I see `cfg(std)`, I know where that `std` configuration is coming from just by looking at the `Cargo.toml`. Just my two cents. Thank you for confirming that it's functionally correct.

jhpratt:
> Boy do I feel silly :flushed:.

No worries! Unless you were looking at who has control on crates.io or who has committed to the repos, there's no reason for you to know. And realistically, it doesn't much matter.

> As I've mentioned, I'm pretty new. I think I've seen it as in my commit more, but don't have strong feelings about it. I think for me, it doesn't _hide_ the `no_std` flag. When I see `cfg(std)`, I know where that `std` configuration is coming from just by looking at the `Cargo.toml`. Just my two cents. Thank you for confirming that it's functionally correct.

For what it's worth if it's clearer to new people, that's typically what's preferred! cfg aliases certainly isn't something a beginner needs to be familiar with, either.

time-macros/src/lib.rs
jhpratt: Looks like it is necessary, as it's not a proc_macro crate.

jhpratt:
Looks like everything is working as expected. All I really did was get CI to a place I like.

Thank you!

PR #250, opened 2020-04-21
add 'alloc' feature, enabling use on targets without a global allocator

berkowski:
add 'alloc' feature, enabled by default by 'std'

Disabling the 'alloc' feature allows use on targets without an global allocator.


berkowski:
## Addresses
Current `#![no_std]` builds still require a global allocator to use the `time` crate.  This PR seeks to remedy that.

## Current Status
Added an `alloc` feature and gated code until it compiled and ran on an STM32L073 (thumbv6m).  This is a heavy-handed approach and needs to be revisited.  For example, some fix-string parsing should be achievable instead of just removing all parsing all together. 

## Open questions/work:
- [ ] How should the `alloc` and `std` features interact
- [ ] Enable fixed-format parsing (e.g. RFC3339)
- [ ] Make a CI target using `--no-default-features` on an ARM  toolchain

jhpratt:
Just a quick note with regard to interaction between `alloc` and `std` features: This is only a problem if MSRV is < 1.36. It's currently 1.34 on the latest release and 1.32 on master. I'm not opposed to raising it in 0.3 if it makes things significantly easier.

If MSRV is ≥ 1.36, then the `std` feature can just rely on `alloc`.

berkowski:
Systems with custom allocators could use `no_std` with `alloc`.  I think there's value in keeping them separate if the `#cfg` statements required don't get out of hand.

jhpratt:
@berkowski Probably best to work from the `v0.3` branch, which is on GitHub now. It has an MSRV of 1.36, so the `std` feature can unconditionally rely on an `alloc` feature.

CI is also a bit more manageable, as type checking was removed — tests can't run without valid type checking, of course.

berkowski:
@jhpratt Thanks, I'll take a look at 0.3 for active work.

I haven't used "draft" PRs before.  I had hoped it would cut down on the notification spam but maybe not?  Would you prefer me working on another private branch until things are closer?

jhpratt:
Up to you! I get plenty of emails per day anyways, so it's not a big deal. If it gets crazy, I can always unsubscribe from the issue and rely on you pinging me :slightly_smiling_face:

berkowski:
@jhpratt how open would you be to using the [`heapless`](https://crates.io/crates/heapless) crate to enable use of `time` without a global allocator?  It may be more straight-forward to implement and could preserve much of the format/parsing capability.

I ask because my rebasing of the original 0.2 code is becoming harder with the 0.3 work and I think I need to start again from scratch anway.

jhpratt:
Feel free to add whatever dependencies initially — we can remove things later on. I don't think it's possible to have it only enabled for no-alloc situations, since cargo doesn't support negative features yet.

PR #258, opened 2020-05-07
Add timestamp serde support via with-annotation

### LINKED ISSUE #255: Serde OffsetDateTime as Unix timestamp
I need to de/serialize `OffsetDateTime` from/to Unix timestamps.

For that purpose I wrote a module for use in combination with [serde's with-annotation](https://serde.rs/field-attrs.html#with).
Would such a module be in scope for this project? If yes, I would happily create a PR.

mwkroening:
Fixes #255.

This adds two modules fulfilling the requirements for serde's with-annotation: `time::serde::timestamp::{self, option}`.

As this requires a public module, I published the `time::serde` module. Would you prefer another place for this instead?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "time"
-version = "0.2.15"
+version = "0.2.16"
 authors = ["Jacob Pratt <the.z.cuber@gmail.com>"]
 edition = "2018"
 repository = "https://github.com/time-rs/time"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ __doc = []
 cfg-if = "0.1.10"
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }
-serde_json = { version = "1", optional = true }
+_serde_json = { package = "serde_json", version = "1", optional = true }
 standback = { version = "0.2.5", default-features = false }
 time-macros = { version = "0.1", path = "time-macros" }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ __doc = []
 cfg-if = "0.1.10"
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }
-_serde_json = { package = "serde_json", version = "1", optional = true }
 standback = { version = "0.2.5", default-features = false }
 time-macros = { version = "0.1", path = "time-macros" }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,4 +50,5 @@ stdweb = { version = "0.4", default-features = false, optional = true }
 version_check = "0.9"
 
 [dev-dependencies]
+serde_json = "1"
 time-macros = { path = "time-macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,12 +25,13 @@ std = ["libc", "winapi", "stdweb", "standback/std"]
 # Internal usage. This is used when building for docs.rs and time-rs.github.io.
 # This feature should never be used by external users. It will likely be
 # removed in the first release after the `doc_cfg` feature gets stabilized.
-__doc = []
+__doc = ["serde_json"]
 
 [dependencies]
 cfg-if = "0.1.10"
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }
+serde_json = { version = "1", optional = true }
 standback = { version = "0.2.5", default-features = false }
 time-macros = { version = "0.1", path = "time-macros" }
 
@@ -50,5 +51,4 @@ stdweb = { version = "0.4", default-features = false, optional = true }
 version_check = "0.9"
 
 [dev-dependencies]
-serde_json = "1"
 time-macros = { path = "time-macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ std = ["libc", "winapi", "stdweb", "standback/std"]
 # Internal usage. This is used when building for docs.rs and time-rs.github.io.
 # This feature should never be used by external users. It will likely be
 # removed in the first release after the `doc_cfg` feature gets stabilized.
-__doc = ["serde_json"]
+__doc = []
 
 [dependencies]
 cfg-if = "0.1.10"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -373,7 +373,7 @@ mod primitive_date_time;
 mod rand;
 #[cfg(serde)]
 #[allow(missing_copy_implementations, missing_debug_implementations)]
-mod serde;
+pub mod serde;
 /// The `Sign` struct and its associated `impl`s.
 mod sign;
 /// The `Time` struct and its associated `impl`s.
--- a/src/serde/timestamp.rs
+++ b/src/serde/timestamp.rs
@@ -1,3 +1,39 @@
+//! De/serialize [`OffsetDateTime`] from/to [Unix timestamps](https://en.wikipedia.org/wiki/Unix_time).
+//!
+//! Use this module in combination with [serde's with-annotation](https://serde.rs/field-attrs.html#with).
+//!
+//! Note that the timestamp represenatioon lacks [`UtcOffset`],
+//! which is being lost on serialization and assumed 0 on deserialization.
+//!
+//! # Examples
+//!
+//! ```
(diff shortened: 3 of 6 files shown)


mwkroening:
Removed a trailing whitespace.

mwkroening:
And fixed another small formatting error.

mwkroening:
I am not sure if that would be equivalent.

Serde's with-annotation requires a module containing a `fn<S>(&T, S) -> Result<S::Ok, S::Error> where S: Serializer` and a `fn<'de, D>(D) -> Result<T, D::Error> where D: Deserializer<'de>` function, which are custom de/serialization implementations. This allows us to de/serialize OffsetDateTime as `i64` instead of a struct or tuple containing `i64` and further allows us to customize the message, that we expect a Unix timestamp, accordingly.

I am not sure if your approach would be possible and I haven't seen it done that way. The [serde-with](https://github.com/jonasbb/serde_with) crate seems to do it manually as well.

mwkroening:
About the remaining checks: I introduced `serde-json` as dev-depencency for the examples. `serde-json` requires `serde` to be built with std: https://serde.rs/no-std.html#no-std-support

I am not sure yet what to do about that.

jhpratt:
@dtolnay ^^ Is that possible? Am I wrong in my thinking of how well interop would be with other crates? I'm not terribly familiar with serde — only the basic usage from an end user's perspective.

@mwkroening With regard to serde-json, could a different crate be used that has `#![no_std]` support, even if limited? If not, it is possible to skip a doctest.

dtolnay:
This seems like the implementation I would have in mind. I am not clear what alternative you have in mind by "secondary struct implementation".

src/serde/timestamp.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Fullfills the requirements for [serde's serialize_with-annotation](https://serde.rs/field-attrs.html#serialize_with).
+    ///
+    /// Prefer using the parent module instead for brevity.
+    pub fn serialize<S>(opt: &Option<OffsetDateTime>, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match *opt {
+            Some(ref value) => super::serialize(value, serializer),
```
dtolnay, line 143: This needs to call serialize_some or some formats won't be able to deserialize the result.

jhpratt:
Alright. If that's the way to do it, then I have faith :stuck_out_tongue:

The way I've been doing serialization/deserialization so far in v0.2 is by using `From`/`Into` impls, which allowed me to alter internal representations with back-compatibility. Obviously that's not an issue here, but I just feel it keeps things cleaner.

I'll take a look at this PR tomorrow or Saturday.

Thanks for the quick response!

src/serde/timestamp.rs
mwkroening: Thanks for the heads up!

mwkroening:
> @mwkroening With regard to serde-json, could a different crate be used that has `#![no_std]` support, even if limited? If not, it is possible to skip a doctest.

There's [serde_json_core](https://japaric.github.io/serde-json-core/serde_json_core/index.html), which is in maintenance mode. Would you prefer using that or disabling the doctests?

jhpratt:
Completely skipping that doctest is fine. If it's not too difficult to add a conditionally-enabled test (not doctest), that would be good to ensure things don't break.

And thanks for that. I'll take a look!

mwkroening:
I instead conditionally enabled the doctest. Is that okay?

jhpratt:
I'm a bit concerned about serde_json being added as a runtime dependency. Is that actually necessary? And given it's for internal usage, I'd prefer it be gated by an underscore-prefixed name.

Sorry for being picky! Just quickly looked things over.

mwkroening:
Having it as a dev-dependency does not work, as it would force std in serde and optional dev-dependencies are not supported yet. That's why I made it an optional regular dependency. I did not find another way to work around that.

I prefixed it with an underscore.

No worries! I think being picky is definitely a good thing here.

mwkroening:
@jhpratt, I think I might have been able to offload most of the manual work to a wrapper struct as you proposed.

I'll still need to test this further and need another opinion on this, but I'll try to commit my proposal tomorrow.

mwkroening:
While we can't avoid implementing de/serialize, most of that work is now being done by a wrapper, as shown here: https://github.com/serde-rs/serde/issues/1301#issuecomment-394108486
A newtype_struct is being represented by the inner value in json ([Structs and enums in JSON · Serde](https://serde.rs/json.html)), but I am not sure, if that holds true for every format. What do you think, @dtolnay?

Aside from that I improved the examples.

mwkroening:
We could also provide an implementation for `Vec<OffsetDateTime>`, but I am not sure if that would be a common use case. What do you think?

jhpratt:
We could use `#[serde(transparent)]` to guarantee that, which would hold for any serializer/deserializer.

I'll take a look at the PR as a whole later today. I should be able to push up any additional commits for minor nits myself (that's the default option for PRs).

And nah, I think `Vec<OffsetDateTime>` is enough of a niche use-case that the user should start doing the work at that point.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ __doc = []
 cfg-if = "0.1.10"
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }
+_serde_json = { package = "serde_json", version = "1", optional = true }
```
jhpratt, line 34: Since optional dev-dependencies aren't possible, I'd prefer this be left out and just ` ```ignore ` the test. Hopefully some day it'll be allowed.

src/lib.rs
```diff
@@ -373,7 +373,7 @@ mod primitive_date_time;
 mod rand;
 #[cfg(serde)]
 #[allow(missing_copy_implementations, missing_debug_implementations)]
-mod serde;
+pub mod serde;
```
jhpratt, line 376: This'll likely cause merging/rebasing issues for v0.3, since the mod was removed, but that's nothing to worry about. I'll just handle the conflict when it arises.

Note to self: be sure to remove the `#[allow]` for v0.3.

src/serde/timestamp.rs
```diff
@@ -0,0 +1,151 @@
...
+//! assert_eq!(v, serde_json::to_value(&s)?);
+//! assert_eq!(s, serde_json::from_value(v)?);
+//! # Ok(())
+//! # }
+//! # test().unwrap();
+//! # }
+//! ```
+//!
+//! [`OffsetDateTime`]: crate::OffsetDateTime
+//! [`UtcOffset`]: crate::UtcOffset
```
jhpratt, line 39: Docs are built on nightly (both for time-rs.github.io and docs.rs) with intra-crate links enabled, so this isn't necessary to specify.

src/serde/timestamp.rs
```diff
@@ -0,0 +1,151 @@
...
+use crate::OffsetDateTime;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// Fullfills the requirements for [serde's serialize_with-annotation](https://serde.rs/field-attrs.html#serialize_with).
+///
+/// Prefer using the parent module instead for brevity.
+#[allow(single_use_lifetimes)]
+pub fn serialize<S>(datetime: &OffsetDateTime, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
```
jhpratt, line 50: Does serde require it be in a `where` clause, or can we do `<S: Serializer>`?

src/serde/timestamp.rs
mwkroening: I removed the explicit link for `OffsetDateTime`, but the one for `UtcOffset` seems to be required, as it is not `use`d.

src/serde/timestamp.rs
mwkroening: I removed the `where` clause on the two `serialize`s. Do you want them removed on the `deserialize`s as well?

jhpratt: Yeah, that would be great.

jhpratt:
Just pushed up some changes to documentation (to avoid exposing what was internal details), in addition to extracting the duplicated structs.

I wrote a quick module to test this out, but if you could check if with your use case before merging, that would be great.

To rely on this branch, use this in your Cargo.toml:

```toml
time = { git = "https://github.com/mwkroening/time", branch = "serde-timestamp" }
```

mwkroening:
I already tested this regularly with my use case and it still works well. Your changes look good to me.

Thanks a lot for your help!

jhpratt:
Awesome. I'll bump the patch version and release tomorrow.

Thanks again for your work!

PR #265, opened 2020-06-27
Change cfg_attr(tarpaulin, skip)

xd009642:
So in 0.13.4 (yanked) tarpaulin actually started passing `--cfg=tarpaulin` so users could do `#[cfg_attr(tarpaulin, ignore)]` to ignore certain tests in coverage and also conditionally include/exclude code in coverage. This had the unintended result of breaking the recommended skip attr. 

Because of that 0.13.4 was yanked and later on a 0.14.0 version released which still did `--cfg=tarpaulin` but also `--cfg=tarpaulin_include` so that code could now be skipped but still included if you did `#[cfg(not(tarpaulin_include))]`. There's also an unstable `#[tarpaulin::skip]` tool attribute for those using nightly.

This PR changes time to use the new attributes as a user of time and tarpaulin found this broke their coverage CI. 

Link to relevant tarpaulin issue: https://github.com/xd009642/tarpaulin/issues/487

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -736,7 +736,7 @@ impl Duration {
 /// Functions that have been renamed or had signatures changed since v0.1. As
 /// such, they are deprecated.
 #[cfg(v01_deprecated_api)]
-#[cfg_attr(tarpaulin, skip)]
+#[cfg(not(tarpaulin_include))]
 #[allow(clippy::missing_docs_in_private_items, clippy::missing_const_for_fn)]
 impl Duration {
     #[inline(always)]
--- a/src/instant.rs
+++ b/src/instant.rs
@@ -123,7 +123,7 @@ impl Instant {
 impl Instant {
     #[inline(always)]
     #[cfg(v01_deprecated_api)]
-    #[cfg_attr(tarpaulin, skip)]
+    #[cfg(not(tarpaulin_include))]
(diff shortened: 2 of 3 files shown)


xd009642:
Also types shouldn't be covered in coverage so I removed the attributes from the type declarations

jhpratt:
Is tarpaulin able to handle methods with `#[inline]` now? I honestly haven't run coverage checks in a very long time, largely due to it being relatively inaccurate because of inlining.

I can certainly check this out when I'm on my laptop to see how much tarpaulin has improved, as I know there has been work on the compiler's end of things. I won't rule out just removing the cfgs outright, as I actually didn't realize they were still in the codebase.

Thank you for the PR, though! I'll get back with my findings as appropriate!

xd009642:
I think inline might still have some issues, though there are some other changes that give a route to fixing them. It's a bit tricky though in terms of source analysis as it starts to move the analysis context across file boundaries

xd009642:
Types should definitely be excluded though, if they're included that would be a regression :slightly_smiling_face:

jhpratt:
Turns out I'd already removed all `skip` attributes on the v0.3 branch, and running `cargo tarpaulin` after installing the most recent release works as expected. As the v0.3 branch is where almost all development is taking place now, I'll close this.

PR #273, opened 2020-08-25
On NetBSD, link tzset to __tzset50.

### LINKED ISSUE #272: Link `tzset` to `__tzset50` for NetBSD
Similar to the `setlocale` method, NetBSD has this weird compatibility layer and really expects applications to link to `__tzset50` instead ([see their `time.h` header](http://cvsweb.netbsd.org/bsdweb.cgi/~checkout~/src/include/time.h?rev=1.47&content-type=text/plain&only_with_tag=MAIN)).

[This is how `libc` does it for `setlocale`](https://docs.rs/libc/0.2.76/src/libc/unix/mod.rs.html#1379):
```rust
extern "C" {
    // ...
    #[cfg_attr(target_os = "netbsd", link_name = "__setlocale50")]
    pub fn setlocale(
        category: ::c_int,
        locale: *const ::c_char,
    ) -> *mut ::c_char;
    // ...
}
```

I think the same should be done for [`tzset` here](https://github.com/time-rs/time/blob/bc2cdc375206a3ba118ffe16eb7262db300a1201/src/utc_offset.rs#L339):

```rust
                extern "C" {
                    #[cfg_attr(target_os = "netbsd", link_name = "__tzset50")]
                    fn tzset();
                }

```

gyscos:
Fixes #272.

### DIFF
--- a/src/utc_offset.rs
+++ b/src/utc_offset.rs
@@ -337,6 +337,7 @@ fn try_local_offset_at(datetime: OffsetDateTime) -> Option<UtcOffset> {
             /// on any error.
             fn timestamp_to_tm(timestamp: i64) -> Option<libc::tm> {
                 extern "C" {
+                    #[cfg_attr(target_os = "netbsd", link_name = "__tzset50")]
                     fn tzset();
                 }
 


jhpratt:
I presume this currently fails on NetBSD? I ask because I see this bit of code in the header:

```c
#ifndef __LIBC12_SOURCE__
void tzset(void) __RENAME(__tzset50);
#endif
```

I only have a basic understanding of C, and no experience with OS-level things in C, but it looks like this should be aliased?

If this patch fixes something that currently doesn't work, then by all means I'll merge! I was planning a small release for 0.2 later this week anyways, so the timing is good.

krytarowski:
This patch is good, please commit.

gyscos:
Yeah what `void tzset(void) __RENAME(__tzset50);` does is, at the source level, redirrect calls to `tzset` to `__tzset50`. But it only works for actual C code that imports `time.h`; the true `tzset` symbol is part of the compatibility layer and doesn't really work.

I agree it'd be nice to have a NetBSD-test that we can show fails before this patch, and passes after. I'm not really a NetBSD user so I'm not 100% sure how to go about that; maybe some CI service offer NetBSD builds?

In https://github.com/hrkfdn/ncspot/pull/256 @krytarowski mentioned building this crate (albeit for an older version) resulted in the following warning:
```
          ld: /tmp/ncspot/target/debug/deps/libtime-166e0e30576a5331.rlib(time-166e0e30576a5331.time.ey6x0hik-cgu.1.rcgu.o): in function `time::tzset':
          /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.1.43/src/lib.rs:287: warning: warning: reference to compatibility tzset(); include <time.h> for correct reference
```

jhpratt:
Thanks for the confirmation @krytarowski. I see you're a member of the NetBSD organization, so your word is as good as anyone's.

@gyscos I'll take a look to see if I can add NetBSD to CI. It wouldn't be testing, but even a build test would be better than nothing.

PR #274, opened 2020-09-07
Fix incorrect parsing of the offset of a RFC-3339 time string

twistedfall:
The offset in 2020-09-08T08:44:31+02:30 was incorrectly interpreted as 2 minutes and 30 seconds.


jhpratt:
Formatting failure is unrelated to this PR.

Thank you for this, and the test too to ensure no regressions!

PR #277, opened 2020-09-15
fixed solaris build

AviKozokin:
Missing std::TryFrom prevented successful build on x86_64-sun-solaris target

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ Versioning].
 
 ---
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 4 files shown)


jhpratt:
Two relatively minor things:

- Would you mind adding Solaris to CI? It should amount to a single line addition if the target has been around long enough.
- If the Solaris target has been around since before `TryFrom` was stabilized, the import should be `use standback::convert::TryFrom`. This allows usage on older compilers, while maintaining forward compatibility.

jhpratt:
Closing in favor of #278, which is now merged.

PR #279, opened 2020-09-28
Added Default support for Instant

genusistimelord:
Added Default Support to Instant so users can use this with the Default Derive.

### DIFF
--- a/src/instant.rs
+++ b/src/instant.rs
@@ -125,6 +125,12 @@ impl Instant {
     }
 }
 
+impl Default for Instant {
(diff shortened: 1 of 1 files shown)


jhpratt:
Given that:

1. this means `Default::default()` returns a different value on each invocation,
1. `std::time::Instant` does not `impl Default`, and
1. the only purpose of this struct is to provide trait implementations not otherwise possible,

I'm going to go ahead and close this PR as rejected. Thank you for the PR, however.

PR #282, opened 2020-10-02
Support wasm32-unknown-unknown target

RogueliteCollector:
Right now time doesn't fully support wasm32-unknown-unknown target because rust's `std::time` doesn't implement `SystemTime::now()` and `Instant::now()` there, and previous implementation(#224 ) requires `cargo_web` and `stdweb` to function.
Replacing `std::Instant` with an implementation from `instant` crate and `SystemTime::now()` in `OffsetDateTime` with JS equivalent seems to fix most of the issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ libc = { version = "0.2", optional = true }
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["minwinbase", "minwindef", "timezoneapi"], optional = true }
 
-[target.'cfg(target_arch = "wasm32")'.dependencies]
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 instant = { version = "0.1.7", features = [ "now", "wasm-bindgen" ] }
 wasm-bindgen = { version = "0.2.68", optional = true }
 js-sys = "0.3.45"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ rustdoc-args = ["--cfg", "__time_02_docs"]
 default = ["deprecated", "std"]
 deprecated = []
 panicking-api = []
-std = ["libc", "winapi", "wasm-bindgen", "standback/std"]
+std = ["libc", "winapi", "wasm-bindgen", "js-sys", "standback/std"]
 
 [dependencies]
 const_fn = "0.4.2"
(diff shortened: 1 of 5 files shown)


RogueliteCollector:
not sure why web build is failing, but `#![no_std]` isn't supported on wasm32-unknown-unknown with wasm-bindgen either way.

jhpratt:
The error on the CI failure seems pretty clear — time 0.1 is being pulled in as a dependency.

I'm on mobile currently, so I'll look through things more thoroughly once on my laptop if you'd like.

Even upon quick inspection, there are some issues (aside from CI). By far the biggest is the `instant` crate — its license (BSD 3-clause) is incompatible with MIT, and as such cannot be a dependency anywhere in the tree. I'm also concerned by the fact that despite re-exporting `std::time::Instant`, it relies on time 0.1 instead of `std::time::SystemTime` to perform a calculation. I'm not sure why that feature flag needs to be enabled, as it doesn't appear to be used anywhere (though I may be missing it because I'm on mobile). Another concern is the sheer number of dependencies — I'm trying to cut down on the size of the crate overall. 

More generally, I believe a change like this should be made somewhere else, be it in stdlib or one of the crates for interoperability with web APIs. It was my intent to expose the internal structure of `time::Instant` as a thin wrapper, as the documentation already implies that's the case.

RogueliteCollector:
Actually, the number of dependencies was what made me look into the implementation of wasm support. Currently wasm32 target depends on `stdweb` create, which pulls in way too many dependencies(including `js-sys` and `wasm-bindgen`) and was last updated a year ago. `instant` is a small crate that implements `Instant::now` for the wasm32-unknown-unknown target, but I guess it could be easily reimplemented. `instant` depending on `time` is strange though - I don't see it used anywhere in the crate.

RogueliteCollector:
Alright, if `time::Instant` isn't supposed to be used externally then there is no need to implement it for wasm. `stdweb` should still be replaced with just `wasm_bindgen` though since current implementation only supports `cargo-web` and doesn't work when building with `wasm-pack`. But doing so breaks the web build test because it requires newer rustc version.

jhpratt:
`time::Instant` is absolutely intended to be used externally, otherwise it wouldn't exist. All I'm saying is that it is a thin wrapper around the standard library equivalent, and I intended to make that field public in 0.3.

wasm-bindgen could presumably have some way to shim `SystemTime`/`Instant`, and I believe that's the appropriate location to do so.

I'll review this when I'm on my laptop (tonight more than likely), You said it breaks MSRV, so this might be something worth basing off the v0.3 branch — MSRV is 1.36 there.

cyqsimon:
Any progress on this?

jhpratt:
@cyqsimon No; the PR has been closed for months. Feel free to review the comments for my concerns (and ensuing lack of activity).

Issue #256, opened 2020-05-07
OffsetDateTime::replace_time?


djahandarie:
I find myself writing code like this:

```rust
my_offset_date_time
  .date()
  .with_time(time!(05:00))
  .assume_offset(my_offset_date_time.offset())
```

which, while is fairly explicit about what is going on, feels a little long compared to

```rust
my_offset_date_time.replace_time(time!(05:00))
```

(I figured `with_time` is not appropriate thing to call this, because the rest of the library uses that for combining things rather than replacing parts of things. I imagine `set_` could also be a decent name, though that very slightly suggests more of a &mut self interface to me, as opposed to just a self)

And in fact, since it seems that "replacing parts of things" is generally not a pattern exposed anywhere in the library, this would likely open up the pandora's box of "but I want to replace the hour of this `Time`" etc. etc., so I could understand why one might not want to go down this route.

That said, I'd find this quite ergonomic, so here's my shot at suggesting it!

jhpratt:
Something along the lines of `set_` might be feasible. I'll have to check on the interop between `Copy` types and `&mut self` methods, as I'm not sure off the to of my head.

I'm definitely open to this in some form, though, as it's a pattern I have thought about previously.

jhpratt:
For those coming here from TWIR, I thought this issue would be a good one to tackle for Hacktoberfest!

What I'd like to see in a PR is a number of methods (on `Time`, `Date`, `PrimitiveDateTime`, and `OffsetDateTime`) with the signature `fn replace_foo(self) -> Self`, _not_ a mutating version as mentioned in my previous comment. All components should be able to be replaced, even if they are not directly present in the struct definition. `OffsetDateTime` replacements should operate similarly to the getters, in that the offset is taken into account before passing it through to the `PrimitiveDateTime`.

As with any change, documentation and tests should be present. Don't worry about what is done when, as I can clean up the commit history before merging.

Note that any PR doing this should be based on the v0.3 branch, not main. v0.3 is relatively stable with the exception of formatting.

Ask-0:
I would like to claim this one.

PR #284, opened 2020-10-07
OffsetDateTime::replace_foo Functions

Ask-0:
#256

### DIFF
--- a/src/offset_date_time.rs
+++ b/src/offset_date_time.rs
@@ -546,6 +546,88 @@ impl OffsetDateTime {
     }
 }
 
+/// Methods that replace part of the `OffsetDateTime`.
+impl OffsetDateTime {
+    /// Replace the time, which is assumed to be in the stored offset. The date
+    /// and offset components are unchanged.
+    ///
+    /// ```rust
+    /// # use time_macros::{datetime, time};
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 5:00 UTC").replace_time(time!("12:00")),
+    ///     datetime!("2020-01-01 12:00 UTC")
+    /// );
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 12:00 -5").replace_time(time!("7:00")),
+    ///     datetime!("2020-01-01 7:00 -5")
+    /// );
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 0:00 +1").replace_time(time!("12:00")),
+    ///     datetime!("2020-01-01 12:00 +1")
+    /// );
+    /// ```
+    #[must_use = "This method does not mutate the original `OffsetDateTime`."]
+    pub fn replace_time(self, time: Time) -> Self {
+        (self.utc_datetime + self.offset.as_duration())
+            .replace_time(time)
+            .assume_offset(self.offset)
+    }
+
+    /// Replace the date, which is assumed to be in the stored offset. The time
+    /// and offset components are unchanged.
+    ///
+    /// ```rust
+    /// # use time_macros::{datetime, date};
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 12:00 UTC").replace_date(date!("2020-01-30")),
+    ///     datetime!("2020-01-30 12:00 UTC")
+    /// );
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 0:00 +1").replace_date(date!("2020-01-30")),
+    ///     datetime!("2020-01-30 0:00 +1")
+    /// );
+    /// ```
+    #[must_use = "This method does not mutate the original `OffsetDateTime`."]
+    pub fn replace_date(self, date: Date) -> Self {
+        (self.utc_datetime + self.offset.as_duration())
+            .replace_date(date)
+            .assume_offset(self.offset)
+    }
+
+    /// Replace the date and time, which are assumed to be in the stored offset.
+    /// The offset component remains unchanged.
+    ///
+    /// ```rust
+    /// # use time_macros::datetime;
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 12:00 UTC").replace_date_time(datetime!("2020-01-30 16:00")),
+    ///     datetime!("2020-01-30 16:00 UTC")
+    /// );
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 12:00 +1").replace_date_time(datetime!("2020-01-30 0:00")),
+    ///     datetime!("2020-01-30 0:00 +1")
+    /// );
+    /// ```
+    #[must_use = "This method does not mutate the original `OffsetDateTime`."]
+    pub fn replace_date_time(self, date_time: PrimitiveDateTime) -> Self {
+        date_time.assume_offset(self.offset)
+    }
+
+    /// Replace the offset. The date and time components remain unchanged.
+    ///
+    /// ```rust
+    /// # use time_macros::{datetime, offset};
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 0:00 UTC").replace_offset(offset!("-5")),
+    ///     datetime!("2020-01-01 0:00 -5")
+    /// );
+    /// ```
+    #[must_use = "This method does not mutate the original `OffsetDateTime`."]
+    pub fn replace_offset(self, offset: UtcOffset) -> Self {
+        self.utc_datetime.assume_offset(offset)
+    }
+}
+
 /// Methods that allow formatting the `OffsetDateTime`.
 impl OffsetDateTime {
     /// Format the `OffsetDateTime` using the provided string.
--- a/src/primitive_date_time.rs
+++ b/src/primitive_date_time.rs
@@ -371,6 +371,37 @@ impl PrimitiveDateTime {
     }
 }
 
+/// Methods that replace part of the `PrimitiveDateTime`.
+impl PrimitiveDateTime {
+    /// Replace the time, preserving the date.
+    ///
+    /// ```rust
+    /// # use time_macros::{datetime, time};
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 17:00").replace_time(time!("5:00")),
+    ///     datetime!("2020-01-01 5:00")
+    /// );
+    /// ```
+    #[must_use = "This method does not mutate the original `PrimitiveDateTime`."]
+    pub const fn replace_time(self, time: Time) -> Self {
+        self.date.with_time(time)
+    }
+
+    /// Replace the date, preserving the time.
+    ///
+    /// ```rust
+    /// # use time_macros::{datetime, date};
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 12:00").replace_date(date!("2020-01-30")),
+    ///     datetime!("2020-01-30 12:00")
+    /// );
+    /// ```
+    #[must_use = "This method does not mutate the original `PrimitiveDateTime`."]
+    pub const fn replace_date(self, date: Date) -> Self {
+        date.with_time(self.time)
+    }
+}
+
 /// Methods that allow formatting the `PrimitiveDateTime`.
 impl PrimitiveDateTime {
     /// Format the `PrimitiveDateTime` using the provided string.
--- a/tests/offset_date_time.rs
+++ b/tests/offset_date_time.rs
@@ -269,6 +269,54 @@ fn nanosecond() {
 }
 
 #[test]
+fn replace_time() {
+    assert_eq!(
+        datetime!("2020-01-01 5:00 UTC").replace_time(time!("12:00")),
+        datetime!("2020-01-01 12:00 UTC")
+    );
+    assert_eq!(
+        datetime!("2020-01-01 12:00 -5").replace_time(time!("7:00")),
+        datetime!("2020-01-01 7:00 -5")
+    );
+    assert_eq!(
+        datetime!("2020-01-01 0:00 +1").replace_time(time!("12:00")),
+        datetime!("2020-01-01 12:00 +1")
+    );
+}
+
+#[test]
+fn replace_date() {
+    assert_eq!(
+        datetime!("2020-01-01 12:00 UTC").replace_date(date!("2020-01-30")),
+        datetime!("2020-01-30 12:00 UTC")
+    );
+    assert_eq!(
+        datetime!("2020-01-01 0:00 +1").replace_date(date!("2020-01-30")),
+        datetime!("2020-01-30 0:00 +1")
+    );
+}
+
+#[test]
+fn replace_date_time() {
+    assert_eq!(
+        datetime!("2020-01-01 12:00 UTC").replace_date_time(datetime!("2020-01-30 16:00")),
+        datetime!("2020-01-30 16:00 UTC")
+    );
+    assert_eq!(
+        datetime!("2020-01-01 12:00 +1").replace_date_time(datetime!("2020-01-30 0:00")),
+        datetime!("2020-01-30 0:00 +1")
+    );
+}
+
+#[test]
+fn replace_offset() {
+    assert_eq!(
+        datetime!("2020-01-01 0:00 UTC").replace_offset(offset!("-5")),
+        datetime!("2020-01-01 0:00 -5")
+    );
+}
+
+#[test]
 fn format() {
     assert_eq!(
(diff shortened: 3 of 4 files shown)


Ask-0:
@jhpratt Sorry for the delay. I spent some time navigating around Rust and this Crate.  I have a few questions regarding the PR.


1. > What I'd like to see in a PR is a number of methods (on Time, Date, PrimitiveDateTime, and OffsetDateTime) 

All new functions will be under `OffsetDateTime` module. Change `self`  respecting to ` Time, Date, PrimitiveDateTime, and OffsetDateTime`.  Do I understand it correctly?  Is there any more struct that needs `replace_foo`?


2. > with the signature fn replace_foo(self) -> Self

I assume it is like `pub fn replace_foo(self, new_foo: Foo) -> Self` ? 
 
3. At what block in the module should I place the new functions? 

4. I used the same calling methods from the issue in my sample commit.  I will try to generate similar methods, (i.e. using existing functions in the module) for the rest of them.

jhpratt:
> @jhpratt Sorry for the delay. I spent some time navigating around Rust and this Crate. I have a few questions regarding the PR.

No worries! I'm working on formatting myself, so there's still a bit before a 0.3 release regardless.

> All new functions will be under `OffsetDateTime` module. Change `self` respecting to `Time`, `Date`, `PrimitiveDateTime`, and `OffsetDateTime`. Do I understand it correctly? Is there any more struct that needs `replace_foo`?

Ideally I'd like to see it for `PrimitiveDateTime` and `OffsetDateTime`; I don't think there's a huge use case for `Date` or `Time`. For `PrimitiveDateTime`, there should be `replace_date` and `replace_time` methods; for `OffsetDateTime` there should be the same in addition to a `replace_offset` (this one I'm not entirely sold on). Note that for `OffsetDateTime` you'll need to adjust for the already-stored offset — the internal `UtcOffset::as_duration` will be helpful here.

> I assume it is like `pub fn replace_foo(self, new_foo: Foo) -> Self` ?

Correct 🙂 

>  1. At what block in the module should I place the new functions?

Honestly, it doesn't really matter. You can create a new one if you'd like, along with a doc-comment indicating that the contained methods replace existing components.

> 2. I used the same calling methods from the issue in my sample commit.  I will try to generate similar methods, (i.e. using existing functions in the module) for the rest of them.

Ok! There's nothing wrong with relying on existing code; that's what it's there for!

Ask-0:
I pushed the replacement functions for Time, Date, PrimitiveDateTime, UtcOffset in offset_date_time.rs. 
Can you just do a quick chcek? After that, I will do the same thing to the PrimitiveDateTime module. 
Quetions:
1. For naming convention, replace_primitive_date_time or replace_primitivedatetime? 
2. I notice all the macros are expecting string. However the documentation is like `date!(2019-01-01)`. It is a little annonying to copy the code that does not compile. Will this be an another issue? 
This is the error I got: 
```
error: expected string
 --> src/main.rs:9:71
  |
9 |     let mut my_offset_date_time = OffsetDateTime::now_utc().to_offset(offset!(-1));
```

jhpratt:
> I pushed the replacement functions for Time, Date, PrimitiveDateTime, UtcOffset in offset_date_time.rs.
> Can you just do a quick chcek? After that, I will do the same thing to the PrimitiveDateTime module.

Keep in mind that the following should succeed:

```rust
use time::{Time, datetime, time, offset};

let original = datetime!("2020-01-01 0:00").assume_offset(offset!("-5"));
let with_replacement = original.replace_time(time!("12:00"));

assert_eq!(original, datetime!("2020-01-01 0:00").assume_offset(offset!("-5")));
assert_eq!(with_replacement, datetime!("2020-01-01 12:00")).assume_offset(offset!("-5")));
```

This ensures that the time provided is in the known offset, not UTC as is currently assumed. Similar arithmetic should be done for `Date` and `PrimitiveDateTime`. This is where you could use `UtcOffset::as_duration` as previously mentioned. Doing something along the lines of `original.time() + original.offset.as_duration()` might work (check the `+`/`-`, I'm not sure here).

> Quetions:
> 
> 1. For naming convention, replace_primitive_date_time or replace_primitivedatetime?

`replace_date_time` would actually suffice.

> 2. I notice all the macros are expecting string. However the documentation is like `date!(2019-01-01)`. It is a little annonying to copy the code that does not compile. Will this be an another issue?

This behavior is different in 0.3. If you build the docs for the v0.3 branch, the documentation should already be updated. The CHANGELOG should be up-to-date as well. The macros shouldn't be used internally, only in doctests (where they need to be imported from the `time_macros` crate).

Ask-0:
Thank you for explaining.
I was using v2.2 doc from online. I has built the v3 locally now. 👏
**Currently:**
original:
```rust
OffsetDateTime {
    utc_datetime: PrimitiveDateTime {
        date: Date {
            year: 2020,
            ordinal: 1,
        },
        time: Time {
            hour: 5,
            minute: 0,
            second: 0,
            nanosecond: 0,
        },
    },
    offset: UtcOffset {
        seconds: -18000,
    },
}
```
replacement:
```rust
OffsetDateTime {
    utc_datetime: PrimitiveDateTime {
        date: Date {
            year: 2020,
            ordinal: 1,
        },
        time: Time {
            hour: 12,
            minute: 0,
            second: 0,
            nanosecond: 0,
        },
    },
    offset: UtcOffset {
        seconds: -18000,
    },
```
This is the output of the `datetime!("2020-01-01 12:00").assume_offset(offset!("-5"))`:
`OffsetDateTime { utc_datetime: PrimitiveDateTime { date: Date { year: 2020, ordinal: 1 }, time: Time { hour: 17, minute: 0, second: 0, nanosecond: 0 } }, offset: UtcOffset { seconds: -18000 } }`
Since it is replacing, I am currently inheriting the `original`'s offset and only update the new time. I might misunderstand the purpose of the `offset`. So while doing `new_time.time() + self.offset.as_duration()`, do I also inherit the `original`'s offset?

jhpratt:
Yes, the offset does not change. The reason for doing the arithmetic is that the datetime stored isn't in the attached offset, but rather the moment in UTC.

jhpratt:
Alright. Generally speaking, this is what should be happening:

1. Convert `self.utc_datetime` to the stored offset. This can be done by adding (I think) `self.offset.as_duration()` to it. (_Now that I think about it, this probably wouldn't be a bad abstraction for me to add in internally._)
1. Call `.replace_foo` on the resulting `PrimitiveDateTime`.
1. Construct a new `OffsetDateTime`. The simplest way to do this is probably by using the internal `OffsetDateTime::new_assuming_offset`.

The reason for this is that replacing the date/time/offset _must_ not change the other two components (as visible to the user), while the internal storage needs to be adjusted for this expectation. The implementations for `PrimitiveDateTime` look correct.

Regarding `OffsetDateTime::replace_offset`, I'm not sure what the use case might be. However, the method can be removed if a use case doesn't pop up before a release. Given that I intend on (at least) an alpha release before the final v0.3 release, this would provide an adequate time period for use cases to be presented.

Ask-0:
Sorry. I have some school projects due in the past few days. I think my code in the repo already satisfy what you have said in your last comment. I will try to write the docs and test case in this weekend.

jhpratt:
No problem! Let me know when you'd like me to take another look.

Ask-0:
Sorry, my school projects just got clogged altogether.  Can you check if the documentation and test cases are robust enough? Thank you.

jhpratt:
No problem. I'll take a look through things, and can make any adjustments necessary. It'll also give me a chance to verify the tests.

Thank you for your work!

jhpratt:
Testing is important, as arithmetic with time is difficult to wrap your head around. While the implementations for `PrimitiveDateTime` are trivial, `OffsetDateTime` is more complicated. It was my goal to mentor you through this process, as it is undoubtedly difficult for beginners. In my experience, it's best to gently guide someone in the right direction, rather than being blunt (though this admittedly varies by person). That was my intent when I created [this comment](https://github.com/time-rs/time/pull/284#issuecomment-706489015), as I'd noticed that the implementations were likely incorrect (though I had not actually tested them). Unfortunately I don't believe my message came across as intended, as you believed that the only thing remaining was documentation and testing. That was largely my fault, and I apologize.

One thing I learned somewhat early on was to think of the edge case. For handling of times, this is frequently interactions around midnight, as the day might needs to be adjusted. An example of where this can be an issue is the following.

```rust
assert_eq!(
    datetime!("2020-01-01 0:00")
        .assume_offset(offset!("+1"))
        .replace_date(date!("2020-01-31")),
    datetime!("2020-01-31 0:00").assume_offset(offset!("+1"))
);
```

Because `OffsetDateTime` stores the datetime in UTC internally, the internal representation is `datetime!("2019-12-31 23:00")`, stored alongside the offset of UTC+1. Even though you're only replacing the date, the _time of day_ is also relevant due to the possible presence of an offset. This datetime should be adjusted to the local offset (in this case UTC+1), which _might_ alter the date component. The date can then be replaced, and the offset reädded. This is the process that should be followed, likely for replacing the `Time` component of `OffsetDateTime` as well.

Hopefully this makes a bit more sense as to _why_ it's necessary, rather than me just laying out steps that should be followed. Let me know if you have any questions!

---

P.S. I know there's been a bit of back and forth and that school can keep you quite busy; I'm a student as well :slightly_smiling_face: Let me know if you'd rather not bother with making the changes, and I can make them as necessary.

Thank you!

Ask-0:
I misunderstood your previous message. 

>1. Convert self.utc_datetime to the stored offset. This can be done by adding (I think) self.offset.as_duration() to it. (Now that I think about it, this probably wouldn't be a bad abstraction for me to add in internally.)

I printed the value after the function is called. I saw the time is adjusted. I thought the above is what happened internally before we call my replacement function.  Now I resolve it as the following and pass the midnight cases for now.
```
let utc_datetime = (self.utc_datetime + self.offset.as_duration());
        return OffsetDateTime::new_assuming_offset(
            utc_datetime.replace_date(new_date),
            self.offset,
        );
```
What edge cases are also common in the time crate? 

Thank you for your mentoring. I am not used to contributing to open source projects. I usually have a comprehensive understanding of the functions from the beginning. When I step into the middle of projects, what are the procedures should I follow? In this project, I tried to read the docs and the source code that needs to be worked on. What general questions should I ask before working on a project?

jhpratt:
> I printed the value after the function is called. I saw the time is adjusted. I thought the above is what happened internally before we call my replacement function. Now I resolve it as the following and pass the midnight cases for now.

Like I said, it was a miscommunication on my end; I should've been clearer.

> What edge cases are also common in the time crate?

Setting aside formatting/parsing, which is being redone, _a lot_ of date/time arithmetic wraps in some form or another. `OffsetDateTime` (and a future `ZonedDateTime` type) also has edge cases that indirectly resolve to this arithmetic due to the fact that the datetime is stored as UTC internally. Some projects (including this one) have some invariants that must be upheld; failure to do this will likely cause incorrect results.

> I am not used to contributing to open source projects. I usually have a comprehensive understanding of the functions from the beginning. When I step into the middle of projects, what are the procedures should I follow? In this project, I tried to read the docs and the source code that needs to be worked on. What general questions should I ask before working on a project?

Admittedly, there are some things I need to document better internally. Probably the simplest thing to do is ask! Most maintainers look highly of someone who is interested in helping, and will do quite a bit to accommodate if they have time. As to what to ask, that'll depend on the project. For something like this, I obviously don't know exactly what your experience is with date/time stuff in the past, but inquiring about the internal representation, any issues you might run into, and things along those lines.

Ask-0:
I corrected the implementation and added more test cases. Can you do a check? I added some unexpected files. I reverted it and did a pull. I am not sure if I had messed up with the branch, the Github Check failed. 

> OffsetDateTime (and a future ZonedDateTime type) also has edge cases that indirectly resolve to this arithmetic due to the fact that the datetime is stored as UTC internally. 

Will the above edge cases affect this PR?

Thank you.

jhpratt:
I'll take a look at things tomorrow. Not a huge deal if some files are screwed up, as changes can always be reverted and/or force-pushed (though this doesn't much matter as I'll clean up history anyways).

As you're not dealing with implementations of arithmetic, the other edge cases I mentioned shouldn't cause an issue.

jhpratt:
Looks like you did in fact pull in reversions of at least one commit already on the v0.3 branch. It looks like the implementations are correct now, though! There's just a couple minor things left that I'll patch up myself when I clean up the commit history.

So upon quick inspection, the only thing left to be done on your end should be undoing whatever commit it is that you pulled in on accident.

Ask-0:
I tried to revert the commits. It should be fixed now, but I am not sure if I did this in the proper way.

jhpratt:
There was quite a bit that was superfluous in the diff, so I did a manual rebase dropping a large number of commits, followed by merging the current v0.3 branch in.

I'll wait for CI to finish, which shouldn't take too long, then review the code and tests again. Would you prefer me give you feedback one more time or make the (likely minimal) changes myself?

src/offset_date_time.rs
```diff
@@ -546,6 +546,110 @@ impl OffsetDateTime {
...
+    ///     datetime!("2020-01-01 5:00")
+    ///         .assume_offset(offset!("UTC"))
+    ///         .replace_time(time!("12:00")),
+    ///     datetime!("2020-01-01 12:00").assume_offset(offset!("UTC"))
+    /// );
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 12:00")
+    ///         .assume_offset(offset!("-5"))
+    ///         .replace_time(time!("7:00")),
+    ///     datetime!("2020-01-01 12:00").assume_offset(offset!("UTC"))
```
jhpratt, line 567: It would be best to avoid changing the offset between the left and right side, as it makes it less clear that the behavior is correct.

src/offset_date_time.rs
```diff
@@ -546,6 +546,110 @@ impl OffsetDateTime {
...
+impl OffsetDateTime {
+    /// Replacing with a new `Time` component. The new `PrimitiveDateTime` value is stored as UTC internally.
+    ///
+    /// The return value is a `Offset_date_time`.
+    ///
+    /// ```rust
+    ///   # use time_macros::{datetime,offset,time};
+    /// assert_eq!(
+    ///     datetime!("2020-01-01 5:00")
+    ///         .assume_offset(offset!("UTC"))
```
jhpratt, line 559: The `datetime!` macro accepts a `UtcOffset` after the time. This is admittedly quite new; I don't believe it accepted anything other than UTC when you created the PR.

src/offset_date_time.rs
```diff
@@ -546,6 +546,110 @@ impl OffsetDateTime {
...
 }
 
+/// Methods that replace the following components `Date`, `Time`, `PrimitiveDateTime`, `Offset`.
+impl OffsetDateTime {
+    /// Replacing with a new `Time` component. The new `PrimitiveDateTime` value is stored as UTC internally.
+    ///
+    /// The return value is a `Offset_date_time`.
+    ///
+    /// ```rust
+    ///   # use time_macros::{datetime,offset,time};
```
jhpratt, line 556:
```suggestion
    /// # use time_macros::{datetime, offset, time};
```

Just some basic formatting to keep things tidy.

jhpratt (changes requested):
Aside from the clippy lints (which are noted in CI), just a few relatively minor changes. I can make them if you'd prefer, as I'll rewrite the history before merging anyways.

Note that my comments are not exhaustive in location — they likely apply in more than one place.

src/offset_date_time.rs
Ask-0: Should 'cargo test' detect this one? What should I do with this one?

Ask-0:
I did a pull, which showing some unmerged paths. Should I add the **modified** one and rm **deleted by us** to resolve the diverging? Or not bother with this. 
```
Unmerged paths:
  (use "git add/rm <file>..." as appropriate to mark resolution)
	both modified:   src/date.rs
	both modified:   src/lib.rs
	both modified:   src/offset_date_time.rs
	both modified:   src/time_mod.rs
	deleted by us:   time-benchmarks/benches/offset_date_time.rs
	both modified:   time-macros-impl/src/datetime.rs
```
If you prefer, you can change these yourself. I have already learned a lot by doing this PR. 👏

jhpratt:
Yeah, that's just because I force-pushed to fix the clobbered history. I'll go ahead and make the final touch-ups.

Thank you for your work! Hopefully it was a worthwhile experience.

jhpratt (approved):

Issue #249, opened 2020-04-21
support #![no_std] targets without global allocation


berkowski:
#246 gets `time` building on `#![no_std]` but the crate still requires the presence of a global allocator for use.  For example, on a bare-metal thumbv6m target compiled with `--no-default-features` one gets:
```
error: no global memory allocator found but one is required; link to std or add `#[global_allocator]` to a static item that implements the GlobalAlloc trait.

error: `#[alloc_error_handler]` function required, but not found

error: aborting due to 2 previous errors
```
This appears to be mostly due to a few `Box<T>`'s and `Vec<T>`'s used in parsing and error handling.

I've been able to `time` running on my target by taking the heavy-handed approach of pushing offending portions of code behind a new `alloc` feature.  I lose

- String parsing
- `Display` implementations
- The helper `time!` and `date!` macros (I think)

A nice surprise is that `serde` support still works!

To me that's well worth the trade-off of having a library that can do calendar math on an embedded platform.  I can do what limited string parsing I need with something like `nom` and use other constructors here.

It would be great if `time` supported this out of the box but I'm not sure how you would want to go about such an invasive change.  I'd be happy to start a PR if there's interest.

jhpratt:
A couple things:

- `Box<T>` is only used in one or two locations, and for the sole purpose of reducing the size of an enum/struct on the stack. This could be changed, but I'm not sure it's worth it. Benchmarks might be necessary.
- `Vec<T>` is used for two purposes:
  - Error _reporting_ — it's used to store what the minimum and/or maximum values are predicated on for a `ComponentRangeError`. I _might_ be able to use a slice reference combined with a lifetime, but the borrow checker might not agree. There's likely some other hacky ways around it as well.
  - Parsing of the formatting strings. I don't think there's any way around this, as none of this is statically knowable.

Note that _fixed_ formats like RFC3339 (present as of 0.2.10) don't run into the issue with parsing formatting strings. Those could theoretically be supported even in no-alloc environments.

The macros shouldn't cause any issue, as those are done fully at compile-time and emit what is essentially an unchecked constructor.

---

I'm actually interested in switching to nom for parsing entirely. The main issue here is compatibility with custom errors, which _should_ have documentation by the time nom 6 is released. This shouldn't have any effects on public APIs, though.

[...]

PR #285, opened 2020-10-12
Add no-alloc support

garasubo:
Solve #249 

Add `alloc` feature to support no_std targets.
Confirmed that the build passed in my no_std application

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Versioning].
 - `OffsetDateTime::replace_date`
 - `OffsetDateTime::replace_date_time`
 - `OffsetDateTime::replace_offset`
+- `#![no_alloc]` support
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,8 @@ rustdoc-args = ["--cfg", "__time_02_docs"]
 default = ["std"]
 macros = ["time-macros"]
 local-offset = ["std", "libc", "winapi"]
-std = ["standback/std"]
+std = ["standback/std", "alloc"]
+alloc = []
 
 [dependencies]
 const_fn = "0.4.2"
--- a/src/date.rs
+++ b/src/date.rs
@@ -17,10 +17,9 @@ use core::{
     time::Duration as StdDuration,
 };
 #[cfg(feature = "std")]
-use core::{
-    cmp::{Ord, PartialOrd},
-    fmt::Display,
-};
+use core::cmp::{Ord, PartialOrd};
+#[cfg(feature = "alloc")]
+use core::fmt::Display;
 
 /// The minimum valid year.
 pub(crate) const MIN_YEAR: i32 = -999_999;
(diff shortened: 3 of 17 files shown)


garasubo:
One concern is in no_std build, there is an warning from Error Enum
```
warning: enum variant is more than three times larger (52 bytes) than the next largest
   --> src/format/parse.rs:66:5
    |
66  |     ComponentOutOfRange(error::ComponentRange),
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> src/lib.rs:141:5
    |
141 |     variant_size_differences
    |     ^^^^^^^^^^^^^^^^^^^^^^^^
```

jhpratt:
Would you mind adding this to CI? This should amount to removing the `alloc` in the list of features to ignore and always be present — this is on a few lines in the workflow file.

The only other thing I have to note (assuming CI passes once it's done) is that it would be best to remove the `Box` entirely. If this weren't the case, even requiring the `alloc` feature via a transitive dependency could break preëxisting code.

With regard to the lint, feel free to silence it, as there's not really anything actionable.

garasubo:
TODO: remove unused import when all features are disabled

garasubo:
Removed all unused imports.

Currently, format module cannot use in no-std mode without alloc because it depends on `String` and `ToString`.

jhpratt:
Would you mind rebasing this? It looks like there's a couple merge conflicts currently. That should also cause the CI to run, since it didn't on the most recent commit (not sure why).

garasubo:
Rebased my branch to the latest v0.3 branch

jhpratt:
Looks like there's still some unused imports in addition to some lints that could be addressed. Feel free to silence the missing `Copy` warning on `Error` (but not `error::Parse`).

garasubo:
> Looks like there's still some unused imports in addition to some lints that could be addressed. 

Which one? In my env, I didn't see any unused imports in `cargo build`, `cargo build --all-features` and `cargo build --no-default-features`

jhpratt:
Take a look at the CI output; it runs the full powerset of features.

garasubo:
Rebased my branch and fixed the unused import issue

jhpratt:
Looks like aside from formatting (`cargo +nightly fmt` will suffice), `Copy` should still be derived on `error::Parse`, with the remaining clippy lints silenced.

garasubo:
Added `Copy` for `parse::Error`

garasubo:
@jhpratt Do you want me to format this? I assumed that I don't have to do it because the original issue said so.
I believe I resolved any other issues.

jhpratt:
When I referenced formatting, I meant the formatting/parsing parts of the time crate, not rustfmt. Sorry for the confusion via ambiguous wording. Either way, I can handle rustfmt when cleaning up history. Looks like there's a couple lints that are still unhandled, as well as an unexpected `#[derive(Copy)]` on the generic `error::Error` type.

I'll likely clean up the commit history and make the remaining changes myself at some point, as my requests for changes frankly don't seem to be going anywhere.

jhpratt:
doc_cfg annotations will need to be added at some point, but I'll likely do that at a later point.

jhpratt (approved):

Issue #286, opened 2020-10-18
[Feature] Provide Arbitrary trait implementation(s)?


emlun:
I've found [`quickcheck`](https://crates.io/crates/quickcheck) to be a quite useful way to write tests. How about providing `Arbitrary` implementations for the types in this library as an optional feature? I'll volunteer to do it. :slightly_smiling_face:

I'm not completely familiar with the Rust ecosystem on this matter, though (I've mostly done the similar thing in Scala) - it looks like there's both `quickcheck`'s own `quickcheck::arbitrary::Arbitrary` trait, and a different `arbitrary::Arbitrary` trait in the [`arbitrary`](https://crates.io/crates/arbitrary) crate, with no apparent relation between the two. I'm not sure which would be better to implement (if not both) - the quickcheck one seems necessary for quickcheck itself, but [`libfuzzer`](https://github.com/rust-fuzz/libfuzzer) uses the one from the `arbitrary` crate. I'm leaning towards starting with the quickcheck one since that enables more fine-grained tests than fuzzing for full-on crashes.

jhpratt:
While I'm aware of quickcheck, I'm not familiar with exactly how it's used. How would implementing the `Arbitrary` trait (behind a flag, of course) be useful to an end user?

emlun:
Implementing `quickcheck::Arbitrary` would allow users to write tests like this:

```rust
use time::Date;
use quickcheck::quickcheck;

struct DateRange {
    from: Date,
    to: Date,
}

impl DateRange {
    fn new(from: Date, to: Date) -> Result<Self, ()> {
        Ok(DateRange { from, to })
    }
}

quickcheck! {
    fn date_range_is_well_defined(from: Date, to: Date) -> bool {
        let r = DateRange::new(from, to);
        if from <= to {
            r.is_ok()
        } else {
            r.is_err()
        }
    }
}
```

Without it you get a trait bound error:

```
error[E0277]: the trait bound `time::date::Date: quickcheck::arbitrary::Arbitrary` is not satisfied
15  | / quickcheck! {
16  | |     fn date_range_is_well_defined(from: Date, to: Date) -> bool {
17  | |         let r = DateRange::new(from, to);
18  | |         if from <= to {
...   |
23  | |     }
24  | | }
    | |_^ the trait `quickcheck::arbitrary::Arbitrary` is not implemented for `time::date::Date`
```

[...]

PR #288, opened 2020-10-25
Implement quickcheck::Arbitrary for data types

emlun:
Alright, here's a go at #286. This adds a `quickcheck-gen` feature (it can't be just `quickcheck` because it also needs to enable `rand` - unless you require the user to also enable `rand` explicitly) and a subcrate `time-quickcheck-tests` that tests that feature.

I've intentionally left `Instant` out since it seems like values of that type aren't intrinsically meaningful on their own. It also seems quite easy to obtain an `Instant` from a `Duration` if needed, and that type does get an `Arbitrary` implementation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,12 @@ rustdoc-args = ["--cfg", "__time_02_docs"]
 default = ["std"]
 macros = ["time-macros"]
 local-offset = ["std", "libc", "winapi"]
+quickcheck = ["quickcheck-dep", "rand", "std"]
 std = ["standback/std"]
 
 [dependencies]
 const_fn = "0.4.2"
+quickcheck-dep = { package = "quickcheck", version = "0.9", optional = true }
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }
 standback = { version = "0.2.5", default-features = false }
--- a/src/quickcheck.rs
+++ b/src/quickcheck.rs
@@ -40,7 +40,7 @@
 use crate::{
     date::{MAX_YEAR, MIN_YEAR},
     util::days_in_year,
-    Date, Duration, PrimitiveDateTime, Time,
+    Date, Duration, PrimitiveDateTime, Time, UtcOffset,
 };
 use core::{cmp, convert::TryInto};
 use quickcheck_dep::{Arbitrary, Gen};
@@ -190,5 +190,21 @@ impl Arbitrary for PrimitiveDateTime {
         let shrunk_time = time.shrink().map(move |time| Self::new(date, time));
 
         Box::new(shrunk_date.chain(shrunk_time))
+    }
+}
+
+impl Arbitrary for UtcOffset {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let size = g
+            .size()
+            .try_into()
+            .unwrap_or(i32::MAX)
+            .clamp(1, 60 * 60 * 24);
+        let offset = g.gen_range(-cmp::max(0, size - 1), size);
+        Self::seconds_unchecked(offset)
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        Box::new(self.as_seconds().shrink().map(Self::seconds_unchecked))
     }
 }
--- a/src/quickcheck.rs
+++ b/src/quickcheck.rs
@@ -36,3 +36,46 @@
 //! An implementation for `Instant` is intentionally omitted since its values
 //! are only meaningful in relation to a [`Duration`], and obtaining an
 //! `Instant` from a [`Duration`] is very simple anyway.
+
+use crate::{
+    date::{MAX_YEAR, MIN_YEAR},
+    util::days_in_year,
+    Date,
+};
+use core::{cmp, convert::TryInto};
+use quickcheck_dep::{Arbitrary, Gen};
+use rand::Rng;
+#[allow(unused_imports)]
+use standback::prelude::*; // assoc_int_consts (1.43)
+
+impl Arbitrary for Date {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let year_size = g.size().try_into().unwrap_or(i32::MAX);
+        let year = if year_size == 0 {
+            0
+        } else {
+            g.gen_range(
+                cmp::max(MIN_YEAR, -year_size),
+                cmp::min(MAX_YEAR, year_size),
+            )
+        };
+
+        let ordinal_size = cmp::min(g.size().try_into().unwrap_or(u16::MAX), days_in_year(year));
+        let ordinal = g.gen_range(1, cmp::max(2, ordinal_size + 1));
+
+        Self::from_yo_unchecked(year, ordinal)
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let (year, ordinal) = self.as_yo();
+
+        let shrunk_year = year
+            .shrink()
+            .flat_map(move |year| Self::from_yo(year, ordinal));
+        let shrunk_ordinal = ordinal
+            .shrink()
+            .flat_map(move |ordinal| Self::from_yo(year, ordinal));
+
+        Box::new(shrunk_year.chain(shrunk_ordinal))
+    }
+}
--- a/src/quickcheck.rs
+++ b/src/quickcheck.rs
@@ -40,7 +40,7 @@
 use crate::{
     date::{MAX_YEAR, MIN_YEAR},
     util::days_in_year,
-    Date, Duration,
+    Date, Duration, Time,
 };
 use core::{cmp, convert::TryInto};
 use quickcheck_dep::{Arbitrary, Gen};
@@ -131,5 +131,48 @@ impl Arbitrary for Duration {
         });
 
         Box::new(shrunk_seconds.chain(shrunk_nanoseconds))
+    }
+}
+
+impl Arbitrary for Time {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let hour = g.gen_range(0, g.size().try_into().unwrap_or(u8::MAX).clamp(1, 24));
+        let minute = g.gen_range(0, g.size().try_into().unwrap_or(u8::MAX).clamp(1, 60));
+        let second = g.gen_range(0, g.size().try_into().unwrap_or(u8::MAX).clamp(1, 60));
+        let nanosecond = g.gen_range(
+            0,
+            g.size()
+                .try_into()
+                .unwrap_or(u32::MAX)
+                .clamp(1, 1_000_000_000),
+        );
+        Self::from_hms_nanos_unchecked(hour, minute, second, nanosecond)
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let hour = self.hour;
+        let minute = self.minute;
+        let second = self.second;
+        let nanosecond = self.nanosecond;
+
+        let shrunk_hour = hour
+            .shrink()
+            .map(move |hour| Self::from_hms_nanos_unchecked(hour, minute, second, nanosecond));
+        let shrunk_minute = minute
+            .shrink()
+            .map(move |minute| Self::from_hms_nanos_unchecked(hour, minute, second, nanosecond));
+        let shrunk_second = second
+            .shrink()
+            .map(move |second| Self::from_hms_nanos_unchecked(hour, minute, second, nanosecond));
+        let shrunk_nanos = nanosecond.shrink().map(move |nanosecond| {
+            Self::from_hms_nanos_unchecked(hour, minute, second, nanosecond)
+        });
+
+        Box::new(
+            shrunk_hour
+                .chain(shrunk_minute)
+                .chain(shrunk_second)
+                .chain(shrunk_nanos),
+        )
     }
 }
--- a/src/quickcheck.rs
+++ b/src/quickcheck.rs
@@ -40,7 +40,7 @@
 use crate::{
     date::{MAX_YEAR, MIN_YEAR},
     util::days_in_year,
-    Date, Duration, PrimitiveDateTime, Time, UtcOffset,
+    Date, Duration, OffsetDateTime, PrimitiveDateTime, Time, UtcOffset,
 };
 use core::{cmp, convert::TryInto};
 use quickcheck_dep::{Arbitrary, Gen};
@@ -206,5 +206,27 @@ impl Arbitrary for UtcOffset {
 
     fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
         Box::new(self.as_seconds().shrink().map(Self::seconds_unchecked))
+    }
+}
+
+impl Arbitrary for OffsetDateTime {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let datetime = PrimitiveDateTime::arbitrary(g);
+        let offset = UtcOffset::arbitrary(g);
+        datetime.assume_offset(offset)
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let datetime = PrimitiveDateTime::new(self.date(), self.time());
+        let offset = self.offset();
+
+        let shrunk_datetime = datetime
+            .shrink()
+            .map(move |datetime| datetime.assume_offset(offset));
+        let shrunk_offset = offset
+            .shrink()
+            .map(move |offset| datetime.assume_offset(offset));
+
+        Box::new(shrunk_datetime.chain(shrunk_offset))
     }
 }
--- a/src/quickcheck.rs
+++ b/src/quickcheck.rs
@@ -40,9 +40,9 @@
 use crate::{
     date::{MAX_YEAR, MIN_YEAR},
     util::days_in_year,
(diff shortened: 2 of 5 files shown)


jhpratt:
I'll take a look at this when I get a chance.

Don't worry about the CI status right now; I had a change on the v0.3 branch that accidentally broke some things.

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+    cmp::{max, min},
+    convert::TryFrom,
+};
+
+impl Arbitrary for Date {
+    fn arbitrary<G>(g: &mut G) -> Self
+    where
+        G: Gen,
+    {
+        let y_size: i32 = i32::try_from(g.size()).unwrap_or(i32::max_value());
```
jhpratt, line 57: Two things:

- I don't think the type annotation is necessary here.
- It should be possible to just do `.try_into()`, given that the compiler knows an `i32` is expected (due to the `i32::max_value()`).

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+        let y: i32 = if y_size == 0 {
```
jhpratt, line 58: Could we avoid single-letter variable names? It generally makes things easier to read.

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+        let s = i64::arbitrary(g);
+        let ns: i32 = g.gen_range(
+            0,
+            max(
+                1,
+                min(
+                    i32::try_from(g.size()).unwrap_or(i32::max_value()),
+                    1_000_000_000,
+                ),
+            ),
```
jhpratt, line 98: A shim for the unstable `clamp` method would be _super_ useful for readability here. I can provide this if desired.

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+        );
+        Self::new(s, ns * (s.signum() as i32))
```
jhpratt, line 100: Directly constructing the `Duration` would be more efficient due to the checking of invariants in the constructor.

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+            u16::try_from(g.size()).unwrap_or(u16::max_value()),
+            days_in_year(y),
+        );
+        let o = g.gen_range(1, max(2, o_size + 1));
+
+        Self::from_yo_unchecked(y, o)
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let self_y = self.year();
```
jhpratt, line 74: This could probably just be called `year`.

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let orig_dt = PrimitiveDateTime::new(self.date(), self.time());
+        let orig_o = self.offset();
+
+        let shrunk_datetime = orig_dt.shrink().map(move |dt| dt.assume_offset(orig_o));
+        let shrunk_offset = orig_o.shrink().map(move |o| orig_dt.assume_offset(o));
+
+        Box::new(shrunk_datetime.chain(shrunk_offset))
+    }
+}
```
jhpratt, line 239: For `OffsetDateTime` — what would a user be expecting to happen during shrinking? I think (read: I haven't checked this) that this effectively shrinks the date, time, and offset, in that order. As the offset affects the calculated date/time, is this what a user would expect? I believe so, but I'm not sure :thinking:

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+            4 => Friday,
+            5 => Saturday,
+            _ => Sunday,
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        match self {
+            Weekday::Monday => Box::new(std::iter::empty()),
+            _ => Box::new(std::iter::once(self.previous())),
```
jhpratt, line 261: Having not tested this, is `shrink` called multiple times? From looking at the other types, it looks like a complete iterator is returned on the first call, where only a single item is returned here.

jhpratt (changes requested):
For the tests, I just quickly glanced over things, as I don't particularly understand what's going on there.

If you rebase on the v0.3 branch, you should be able to trivially add `quickcheck-gen` to the `std,rand` group, which would add it to CI without doubling the number of checks run.

One thought about the license files: Should the author name be changed to "Jacob Pratt et al."? That would (correctly) indicate that I'm not the sole copyright owner.

Overall, things look good!

emlun:
> If you rebase on the v0.3 branch, you should be able to trivially add `quickcheck-gen` to the `std,rand` group, which would add it to CI without doubling the number of checks run.

I'm not sure what you mean, I don't see anything like that in `Cargo.toml`. Maybe I'm not familiar enough with Cargo...

> One thought about the license files: Should the author name be changed to "Jacob Pratt et al."? That would (correctly) indicate that I'm not the sole copyright owner.

Sounds like a reasonable idea. Do you also want to do something to the `authors` fields in the `Cargo.toml` files? I don't think my contributions here are enough for me to be considered an author of the library proper, just wondering your thoughts on the `time-benchmarks/Cargo.toml` file for example.

> Overall, things look good!

Thanks!

src/quickcheck.rs
emlun: Yes, quickcheck uses `shrink` to iteratively minimize the test case, and it calls the method multiple times while doing so. I'm honestly not sure whether it's better to return "long" iterators at a time or just one, or how much it matters. I considered using `iter::successors` here but ultimately decided against it for no particular reason.

Cargo.toml
```diff
@@ -22,17 +22,19 @@ rustdoc-args = ["--cfg", "__time_02_docs"]
...
 const_fn = "0.4.2"
+quickcheck = { version = "0.9", optional = true }
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1", optional = true, default-features = false, features = ["derive"] }
 standback = { version = "0.2.5", default-features = false }
 time-macros = { version = "0.1", path = "time-macros", optional = true }
 
 [workspace]
-members = ["time-benchmarks", "time-macros", "time-macros-impl"]
+members = ["time-benchmarks", "time-macros", "time-macros-impl", "time-quickcheck-tests"]
```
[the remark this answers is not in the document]
emlun: My main reason was to ensure that the `quickcheck-gen` feature is enabled, so it's impossible to accidentally not run the tests by not having the feature enabled. But I guess that doesn't really matter much since you run the tests against all the features in CI, so I can merge this into the base library if you prefer.

src/quickcheck.rs
emlun: Yep, good catch!

src/quickcheck.rs
```diff
@@ -0,0 +1,264 @@
...
+use rand::Rng;
+use std::{
+    cmp::{max, min},
+    convert::TryFrom,
+};
+
+impl Arbitrary for Date {
+    fn arbitrary<G>(g: &mut G) -> Self
+    where
+        G: Gen,
```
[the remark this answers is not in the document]
emlun: It is, but I personally prefer extracting the `where` clause to make the signature definition less busy. Do you prefer inlining the constraint?

src/quickcheck.rs
emlun: Sure. Should I do the same for the `Gen` parameters?

src/quickcheck.rs
emlun: Heheh, yeah, good idea! I can add one.

Cargo.toml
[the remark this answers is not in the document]
jhpratt: I've already got some other things gated in the tests, so gating this would be in line with that. Obviously it wouldn't be run unless the flag is enabled, but that's to be expected.

src/quickcheck.rs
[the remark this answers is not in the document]
jhpratt: Yeah, I prefer it inlined. I believe that's actually how rustdoc shows it regardless.

src/quickcheck.rs
jhpratt: Nah, since that's a "standard" parameter, if that makes sense. Sorta similar to using `T` for a type parameter.

src/quickcheck.rs
jhpratt: If performance somehow becomes an issue with this, then it can be handled at that point. As long as the behavior is correct, it doesn't matter currently.

jhpratt:
> > If you rebase on the v0.3 branch, you should be able to trivially add `quickcheck-gen` to the `std,rand` group, which would add it to CI without doubling the number of checks run.
> 
> I'm not sure what you mean, I don't see anything like that in `Cargo.toml`. Maybe I'm not familiar enough with Cargo...

This is in the `.github/workflows/build.yaml` file. There are a couple lines with `--group-features` there; I think it'll make more sense once you see that.

> > One thought about the license files: Should the author name be changed to "Jacob Pratt et al."? That would (correctly) indicate that I'm not the sole copyright owner.
> 
> Sounds like a reasonable idea. Do you also want to do something to the `authors` fields in the `Cargo.toml` files? I don't think my contributions here are enough for me to be considered an author of the library proper, just wondering your thoughts on the `time-benchmarks/Cargo.toml` file for example.

Adding people to the "authors" fields is presumably something that you just know when it's the time to do. Though adding a generic "time contributors" is never a bad idea.

With regard to the benchmarks, I believe you're already listed there!

src/quickcheck.rs
emlun: Yeah, I chose to use the constructor to prevent accidentally broken invariants. :slightly_smiling_face: But I'll change it.

src/quickcheck.rs
emlun: Yep, agreed.

emlun:
> > > If you rebase on the v0.3 branch, you should be able to trivially add `quickcheck-gen` to the `std,rand` group, which would add it to CI without doubling the number of checks run.
>
> This is in the `.github/workflows/build.yaml` file. There are a couple lines with `--group-features` there; I think it'll make more sense once you see that.

Oh, ok, did you mean the `serde,rand` group?

> With regard to the benchmarks, I believe you're already listed there!

Yeah, since I did write all of that sub-crate, I took the liberty of claiming authorship of that one. ^^ But I figured while we're on the subject I'd ask your thoughts on it. I don't know how likely it is that I'll remain the "main" author as it continues evolving in the future, but I guess it's just a matter of adding new ones when it seems appropriate.

src/quickcheck.rs
jhpratt: Not an issue. But in cases like this where invariants are trivially proven, we may as well bypass the checks.

jhpratt:
> Oh, ok, did you mean the serde,rand group?

Yeah, just looked over the previous comment and I'm not sure why I put that.

> Yeah, since I did write all of that sub-crate, I took the liberty of claiming authorship of that one. ^^ But I figured while we're on the subject I'd ask your thoughts on it. I don't know how likely it is that I'll remain the "main" author as it continues evolving in the future, but I guess it's just a matter of adding new ones when it seems appropriate.

Absolutely! If it ever truly matters (which realistically is unlikely), I'd just switch the order of the names. As it's not published on crates.io, it's less of an issue.

src/quickcheck.rs
emlun: Yes, I believe it happens in that order - at least I had some issues in another project while implementing shrinking for a `BigRational` type. I had first shrunk the numerator and then the denominator (or maybe vice versa), which caused some issues when one of the two ended up many orders of magnitude greater than the other; reworking the shrinker to effectively alternate between them solved that issue.

My thinking here was to shrink the components in descending order of significance to traverse the value space efficiently, but I'm honestly not sure what a user would expect "shrinking a date" to do exactly. But I guess what matters most is the purpose of shrinking: to find minimal counterexamples. In that light I probably would as a user like to know if the particular offset is significant for the test failure (because it would be shrunk to zero if not). Ultimately I don't think the shrinking order matters much - you'll want the shrinker to generate likely continued failures first, but I don't think any of the components is more or less likely to be significant for any particular test failure.

Maybe it _would_ be more efficient to only generate one shrunk value per component per shrink step, though, so the shrinker doesn't have to run through N shrunk years and N shrunk days in each step before getting to the shrunk offsets, though. That would ultimately depend on the behaviour of `i32::shrink` etc., but I guess we could just limit that using `Iterator::take` anyway.

src/quickcheck.rs
emlun: Done. I left the closure parameters as is, though, as those are extremely local (usually within one line). Let me know if you'd like me to change those too.

src/quickcheck.rs
emlun: I ended up changing it before I saw your response. :innocent:

src/quickcheck.rs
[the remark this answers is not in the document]
emlun: Done.

Cargo.toml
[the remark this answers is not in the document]
emlun: Done!

src/quickcheck.rs
emlun: I see your thumbs up, does that mean you're happy with this as is or would you like me to limit the iterator sizes like I said in the last paragraph? :slightly_smiling_face:

jhpratt: Yeah, it's fine as-is.

src/util.rs
```diff
@@ -86,3 +86,18 @@ pub const fn weeks_in_year(year: i32) -> u8 {
...
+    ///
+    /// If `self` is less than `min`, returns `min`.
+    /// If `self` is greater than `max`, returns `max`.
+    /// Otherwise, returns `self`.
+    fn clamp(self, min: Self, max: Self) -> Self;
+}
+
+impl<T: Ord> Clamp for T {
+    fn clamp(self, min: Self, max: Self) -> Self {
+        std::cmp::max(min, std::cmp::min(self, max))
```
jhpratt, line 101: This should use `core` rather than `std`, and is why CI is failing.

src/util.rs
```diff
@@ -86,3 +86,18 @@ pub const fn weeks_in_year(year: i32) -> u8 {
         52
     }
 }
+
+pub(crate) trait Clamp {
```
jhpratt, line 90: As this isn't used anywhere aside from the quickcheck module, this trait can be placed there (with private visibility).

tests/quickcheck.rs
```diff
@@ -0,0 +1,302 @@
+#![cfg(feature = "quickcheck")]
```
jhpratt, line 1: Is quickcheck `#![no_std]` compatible? If not, the reliance on `std` should be included `Cargo.toml` as well as here.

jhpratt (changes requested):
Quick check looks good aside from a couple nits.

src/util.rs
emlun: Done!

src/util.rs
emlun: Done!

tests/quickcheck.rs
emlun: Since the `shrink` method returns a `std::boxed::Box`, I'm gonna say no. Fixed, thanks!

jhpratt:
Also, just ran across [this post](https://blog.turbo.fish/cargo-features/) on Reddit. Perhaps it may be possible to use the `quickcheck` feature name after all, so long as the internal name is changed (which can ironically be changed right back with a simple `use` statement).

emlun:
Nice find! That does indeed seem to work quite fine.

src/quickcheck.rs
```diff
@@ -0,0 +1,259 @@
...
+            .map(move |o| orig_datetime.assume_offset(o));
+
+        Box::new(shrunk_datetime.chain(shrunk_offset))
+    }
+}
+
+impl Arbitrary for Weekday {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        use Weekday::*;
+        match g.gen_range(0, max(1, min(7, g.size()))) {
```
jhpratt, line 242:
```suggestion
        match g.gen_range(0, g.size().clamp(1, 7)) {
```

jhpratt (commented):
These are minor enough that I can make these changes while cleaning up history. One is just a missed use of `clamp`, while the other is a complete hack that may or may not work.

src/quickcheck.rs
emlun: Oh, good catch! Guess I missed that one because it wasn't spanning 5 lines or whatever... :laughing:

jhpratt:
One question as I'm just about done with the history and verification: The implementation for `UtcOffset` appears to be able to generate `UtcOffset { seconds: -86_400 }` as values. Am I correct in this reading? If so, removing the `- 1` on the generation of the range would make this not the case, correct?

jhpratt:
Ping @emlun — it's just the question in my previous comment that is blocking merging this.

emlun:
Oh, sorry, I didn't notice. I'd say no, since the subtraction happens before the negation, so with `size = 86_400` the lower bound would be `-(86_400 - 1) = -86399`.

However, there _is_ indeed another bug here: with `size = 0`, the range ends up being `[1, 0)` which is invalid and causes a panic. I'll fix that.

jhpratt:
I figured you just didn't see it — no big deal. I've just cleaned up history again, and will merge this pending CI.

jhpratt (approved):

emlun:
Great, thanks! :smile:

PR #290, opened 2020-10-31
Add more cross-links throughout documentation

emlun:
Hopefully the title says it all. :slightly_smiling_face: I saw that [intra-rustdoc links](https://github.com/rust-lang/rust/issues/43466) are used in [some cases](https://github.com/time-rs/time/blob/ee141a366a7cd9b293cbda42a79c0a1459e79178/src/lib.rs#L17) but [not all](https://github.com/time-rs/time/blob/ee141a366a7cd9b293cbda42a79c0a1459e79178/src/date.rs#L604), so I "upgraded" all the ones I could find where it seems appropriate. I did not add links to references to the containing type, for example in "Attempt to create a `Date` from the year, month, and day." on `Date::from_ymd`, since that seems a bit redundant. I also did not rewrap any lines as I'm not sure about the convention on that here, but I'd be happy to do so if desired.

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -594,9 +594,9 @@ impl Date {
     }
 }
 
-/// Methods to add a `Time` component, resulting in a `PrimitiveDateTime`.
+/// Methods to add a [`Time`] component, resulting in a [`PrimitiveDateTime`].
 impl Date {
-    /// Create a `PrimitiveDateTime` using the existing date. The `Time` component will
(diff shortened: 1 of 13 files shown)


jhpratt:
I'll take a look at this tomorrow, but there's no reason to _not_ do this. With regard to comment wrapping, I generally cap the line length to 80 (including indentation). Unfortunately rustfmt breaks  the markdown table in the crate-level documentation, otherwise I'd hand control over to that.

jhpratt (approved):

PR #291, opened 2020-11-09
Remove unused `syn/derive` feature

jyn514:
It was unused. This cuts down build times from 7.9 to 7.2 seconds.
Before: https://jyn514.github.io/assets/cargo-timing-with-derive.html
After: https://jyn514.github.io/assets/cargo-timing-no-derive.html (currently having trouble with my website, but hopefully will be up soon).

### DIFF
--- a/time-macros-impl/Cargo.toml
+++ b/time-macros-impl/Cargo.toml
@@ -12,8 +12,7 @@ description = "Procedural macros for the time crate."
 
 [dependencies]
 proc-macro2 = "1"
-# I'm not entirely sure why `derive` is necessary here, but it fails without it.
-syn = { version = "1.0.15", default-features = false, features = ["proc-macro", "parsing", "printing", "derive"] }
+syn = { version = "1.0.15", default-features = false, features = ["proc-macro", "parsing", "printing"] }
 quote = "1"
 proc-macro-hack = "0.5"
 standback = { version = "0.2", default-features = false }


jhpratt:
IIRC some older versions failed without this present, not that that's a reason not to upgrade.

With regard to the change itself, macros have been redone for v0.3, and will accept a string literal that is then parsed. The only major change to be made before an alpha release is formatting & parsing (the former of which I actually have done to an extent locally). Assuming I don't run into unforeseen issues with testing, I'd expect said alpha release to be within the next couple weeks (no promises, though).

jyn514:
> With regard to the change itself, macros have been redone for v0.3, and will accept a string literal that is then parsed.

Wait, so 0.3 won't have `syn` at all? That's great news :heart: 

Do you think this is still worth backporting to the 0.2 branch?

jhpratt:
That's correct — time 0.3 will not require syn, quote, or proc-macro2. The only time they'll even be in the dependency tree is when an optional feature (like serde) needs it.

With regards to backporting, I'll consider it. There are a number of changes in 0.3 that I could _slightly_ modify in 0.2, though most of those have already been done.

jhpratt:
As I'm quickly going through older issues to tag them correctly, I'd just like to mention that I recently removed the dependency on `serde_derive`, so syn, quote, and proc-macro2 are nowhere to be found in the dependency tree, even with all features enabled! The manual implementation for serialization and deserialization is surprisingly easy once you know what you're doing.

Issue #293, opened 2020-11-10
The call to `localtime_r` may be unsound


quininer:
because `getenv` and `setenv` are not thread-safe, `localtime_r` in time may data race with `std::env::set_var` in libstd.

I described this problem in [chrono issue](https://github.com/chronotope/chrono/issues/499), and time is also affected.

jhpratt:
First off: thanks! It's a nice find.

What would be an appropriate fix for this? 

Mainly for myself: [this](https://github.com/time-rs/time/blob/main/src/utc_offset.rs#L355) is the sole location this function is called.

quininer:
I don't think there is any quick solution, we may have to reimplement `localtime_r` in Rust.

jhpratt:
Ping @ymtvx, who created the original implementation. I'm not terribly familiar with these C APIs.

What output are you expecting from the snippet provided?

quininer:
@jhpratt Oh, I realized that I paste wrong code and it has been updated. POC code will crash.

quininer:
![Screenshot_20201111_111951](https://user-images.githubusercontent.com/6286761/98761197-e4ffa280-240f-11eb-87c0-5045d37ba2e2.png)

quininer:
In addition, I found that when using `time` crate, this crash is easy to reproduce in `glibc`, but it is difficult to use `chrono`.

jhpratt:
So what is _actually_ causing the segfault? Just trying to determine the true root cause so that I can figure out what action to take.

quininer:
`setenv` and `getenv` are not thread-safe, and `localtime_r` will call `getenv`, which cause to data races if `setenv` and `localtime_r` occur at same time.

To be more specific, `getenv` will return a pointer, and after `setenv` the pointer may point to an invalid address.

Rust's libstd avoids this problem by locking `std::env::set_var` and `std::env::var`, but call `localtime_r` directly will bypass the lock.

dekellum:
Rust libstd has a similar, long open issue, for its own calls to `getaddrinfo`: rust-lang/rust#27970

Like `localtime_r`(3),  `getaddrinfo`(3) is similarly marked as vulnerable to concurrent ENV mutation. 

In the linked issue, it is contemplated using the libstd internal ENV lock around calls to `getaddrinfo`, but with concern over performance implications.

hlavaatch:
[...]

Issue #296, opened 2020-11-24
Time 0.2.23 fails to determine local offset


puetzp:
Hey there,

after upgrading to v0.2.23 I am unable to retrieve the correct local offset from my system.
```
println!("{}", UtcOffset::current_local_offset());
```
returns "+0", while
```
println!("{}", UtcOffset::try_current_local_offset());
```
returns a ```IndeterminateOffsetError```.

After downgrading to 0.2.22, both functions correctly return "+1" (CET).

```
date '+%z'
```
also returns "+0100", as expected.

Rust version: rustc 1.47.0 (18bf6b4f0 2020-10-07)

jhpratt:
See #293 and the associated [security advisory](https://github.com/time-rs/time/security/advisories/GHSA-wcg3-cvx6-7396). I've also filed an advisory with RUSTSEC, and this has been assigned a CVE. Not sure what else I could do to make it have more visibility.

Obtaining the systems offset was not and is never guaranteed. It is _always_ best effort.

quininer:
maybe we should provide a feature gate?

jhpratt:
Given that it can outright crash the program with no possible way to recover, I think even a feature gate is a bad idea. That would encourage libraries to use it, and libraries of course have no knowledge of what else their users may be writing.

dekellum:
When you backported 01513cbb in f153a1ca5f [(diff)](https://github.com/time-rs/time/commit/f153a1ca5fdfec979f16c49619e6034cc67e186d#diff-3d5e91be99659fc31172ac5678a87adfb00db42231199ae4eb60c001481803ecR342) and released it as 0.2.23, you "fixed" #293 by essentially disabling the feature of local TZ detection, correct?   

I am surprised by this approach. Should any use of `getaddrinfo` (see rust-lang/rust#27970) also be removed from the rust std library, in your view?  IMO, no. Any consumer of libc/musl/* in any language has the same issue. If you ask the libc maintainers to "fix" this, they'll tell you that the safety constraints are well documented in the man pages, and if you mutate the env in a threaded program, you are responsible for the unsafety. Said another way, the workaround for this footgun is "don't do that."

Shouldn't this (#296) be re-opened since removing a feature (per above) isn't PATCH release compatible?

Or shouldn't this be re-opened, re-titled or replaced with a new open issue because you at least want to re-introduce the local TZ feature in a release like 0.3.0?

[...]

PR #297, opened 2020-12-18
Document #293 in local-offset feature description

jcdyer:
This PR adds documentation to the feature flag for local-offset indicating that the feature is useless on unix platforms, with a reference to #293.

I just spent longer than I wanted trying to figure out why `OffsetDateTime::try_local_now` was returning IndeterminateDateTime.  The path I had to take to figure it out:  

1. Go to the documentation for `OffsetDateTime::try_now_local`.  No mention of the issue.
2. Check the source for that method. No mention of the issue.
3. Go to the documentation for `UtcOffset::local_offset_at`.  No mention.
4. Find the undocumented private function `try_local_offset_at`.  Reference to #293.
5. Check github for #293.  Issue is marked resolved.  
6. Switch dependency to main.  Discover feature flag and new method names.  Error persists.
7. Repeat 1-4 with new nomenclature.  Same reference.

Hopefully a mention on the documentation for the feature flag will prevent others from having to follow that same path.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,7 +30,8 @@
 //! - `local-offset` (_implicitly enables `std`_)
 //!
 //!   This feature enables a number of methods that allow obtaining the system's
-//!   UTC offset.
+//!   UTC offset.  The methods enabled by this feature currently always return
+//!   Err results on unix platforms.  See #293 for details.
 //!
 //! - `large-dates`
 //!


jcdyer:
I'd be happy to also add documentation to the same effect directly on all the `#[cfg(feature = "local-offset")]` methods.

dekellum:
See also #296.  IMO, thats the one that shouldn't have been closed.

jhpratt:
To be clear: #293 is technically resolved. The call is no longer unsound, and will no longer randomly crash your program. #296 was more of a question, and was closed accordingly.

In will look at this PR later today. Adding a note in documentation has been planned; I've just been preoccupied.

jcdyer:
I can update this to point to 296.  The comment in the code mentions 293, so I followed suit.

jhpratt:
Looks like I already added the note to documentation in 288c942a34e19fa38704bdde5e1d49fc21318be2. Unless there's something I missed, I believe this should be closed. The feature flag itself shouldn't need to be documented as such, as it's the individual methods that people will be looking at primarily.

PR #299, opened 2020-12-19
Bump rand to 0.8.0

### LINKED ISSUE #298: Upgrading to rand 0.8
A new version of the [`rand`](https://github.com/rust-random/rand/blob/master/CHANGELOG.md#080---2020-12-18) crate has been released. While the changes needed to upgrade `time` itself are trivial, `quickcheck` still depends on `rand` 0.7.3 and it is unclear when a new version will be released (although there's already a PR at BurntSushi/quickcheck#264).

It is technically possible to rely on both 0.7.3 (for `quickcheck`) and 0.8.0 (for everything else) by doing something like this:
```toml
[dependencies]
rand = { version = "0.8.0", optional = true, default-features = false }
rand-07 = { package = "rand", version = "0.7.3", optional = true, default-features = false }

[features]
quickcheck = ["quickcheck-dep", "rand-07", "std"]
```
Although it's not very pretty. I'll submit a PR to give a better idea of what the changes look like.

MSRV is not affected, since `rand` 0.8 supports Rust >=1.36.

dmit:
Closes #298.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ large-dates = ["time-macros/large-dates"]
 
 [dependencies]
 const_fn = { version = "0.4.4", optional = true }
-quickcheck-dep = { package = "quickcheck", version = "0.9.2", optional = true }
-rand = { version = "0.7.3", optional = true, default-features = false }
+quickcheck-dep = { package = "quickcheck", version = "1.0.1", optional = true }
+rand = { version = "0.8.1", optional = true, default-features = false }
 serde = { version = "1.0.117", optional = true, default-features = false, features = ["derive"] }
 time-macros = { version = "=0.2.0-dev", path = "time-macros", optional = true }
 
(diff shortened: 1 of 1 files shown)


jhpratt:
Going to hold off on this until BurntSushi/quickcheck#264 is merged. I'd prefer to have two versions of the same crate in dependency tree for obvious reasons.

In the future, feel free to just create a pull request for something like this! Issues are only really necessary for larger features.

jhpratt:
Looks like quickcheck has updated to rand 0.8, and is nearing a 1.0 release. When that happens, we'll be all set! I just need to clarify the MSRV policy, as quickcheck requires 1.46 now (time won't bump MSRV to that until mid-Q1 next year).

dmit:
@jhpratt Let me know if you want me to update this PR after the next `quickcheck` release lands, or if you want this closed and to follow a more MSRV-friendly approach.

jhpratt:
Feel free to update it on the next quickcheck release. The MSRV policy will be clarified that any feature enabling an external crate isn't included (unless said feature is enabled by default, which isn't the case anywhere).

jhpratt:
@dmit Would you like to perform the changes, or would you rather I do it? I don't particularly care either way.

jhpratt:
Closing in favor of #304.

PR #301, opened 2021-01-08
Correct Unix timestamp implementation

otake84:
OffsetDateTime::unix_timestamp() in v0.2 is returning different results from v0.3 and chrono's DateTime::timestamp(), so I backported it from v0.3.

See below.

## chrono
```rust
Utc.timestamp(0, 0).timestamp(); // 0
(Utc.timestamp(0, 0) - Duration::nanoseconds(1)).timestamp(); // -1
```

## time v0.3
```rust
OffsetDateTime::unix_epoch().unix_timestamp(); // 0
(OffsetDateTime::unix_epoch() - 1.nanoseconds()).unix_timestamp(); // -1
```

## time v0.2
```rust
OffsetDateTime::unix_epoch().unix_timestamp(); // 0
(OffsetDateTime::unix_epoch() - 1.nanoseconds()).unix_timestamp(); // 0
```

### DIFF
--- a/src/offset_date_time.rs
+++ b/src/offset_date_time.rs
@@ -298,7 +298,13 @@ impl OffsetDateTime {
     /// );
     /// ```
     pub fn unix_timestamp(self) -> i64 {
-        (self - Self::unix_epoch()).whole_seconds()
+        let days = (self.utc_datetime.date.julian_day()
+            - internals::Date::from_yo_unchecked(1970, 1).julian_day())
+            * 86_400;
+        let hours = self.utc_datetime.hour() as i64 * 3_600;
+        let minutes = self.utc_datetime.minute() as i64 * 60;
+        let seconds = self.utc_datetime.second() as i64;
+        days + hours + minutes + seconds
     }
 
     /// Get the [Unix timestamp](https://en.wikipedia.org/wiki/Unix_time).
--- a/src/primitive_date_time.rs
+++ b/src/primitive_date_time.rs
@@ -132,7 +132,13 @@ impl PrimitiveDateTime {
     #[allow(deprecated)]
     #[deprecated(since = "0.2.7", note = "This method assumes an offset of UTC.")]
     pub fn timestamp(self) -> i64 {
-        (self - Self::unix_epoch()).whole_seconds()
+        let days = (self.date.julian_day()
+            - internals::Date::from_yo_unchecked(1970, 1).julian_day())
+            * 86_400;
+        let hours = self.hour() as i64 * 3_600;
+        let minutes = self.minute() as i64 * 60;
+        let seconds = self.second() as i64;
+        days + hours + minutes + seconds
     }
 
     /// Get the `Date` component of the `PrimitiveDateTime`.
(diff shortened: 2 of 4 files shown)


jhpratt:
Looks like this just affects subsecond values for moments before 1970-01-01. I'll look at the semantics when I'm on my laptop, namely the interaction between this and `timestamp_nanos`.

Realistically, I suspect this will have little impact either way, but it's worth checking.

otake84:
I'm working on a new serialization format, and I'm referring to MessagePack for the timestamp specification, but I need to branch the process with `if timestamp >> 34 == 0`, and I'm having trouble serializing the wrong value with the v0.2 behavior.

Please refer to the following.

https://github.com/otake84/messagepack-rs/blob/ceaf01318383b00ccc93e95f065b8fd0afcc8d68/messagepack-rs/src/serializable.rs#L367-L392
https://github.com/otake84/messagepack-rs/blob/ceaf01318383b00ccc93e95f065b8fd0afcc8d68/messagepack-rs/src/deserializable.rs#L106-L118

I hope you will consider this.

jhpratt:
> I'm having trouble serializing the wrong value

In all seriousness, if you're going to say it's the wrong value, please provide a _reason_ why that's the case; it saves me some thinking :slightly_smiling_face: 

After thinking myself, this change in behavior is proper. Without the change, there would be 1.999_999_998 seconds with a Unix timestamp of zero, rather than the 0.999_999_999 seconds for every other timestamp. As such, I am considering this a bug fix.

Could you please make a similar change to the `PrimitiveDateTime::timestamp` method? While deprecated, that's no reason to let it return a value that is not correct.

otake84:
I'm not very good at English, so I may not have communicated well, and I'm very sorry if I offended you...

I forgot to fix PrimitiveDateTime, so I made an additional commit.

This is a MessagePack spec, but it serializes nanoseconds and seconds separately.

If you use the same method as the MessagePack reference implementation, you can compare them with `datetime.unix_timestamp() >> 34 == 0`, and if true, you can serialize them as `timestamp 32` or `timestamp 64`, and if false, as `timestamp 96`. 

In the current v0.2, `(OffsetDateTime::unix_epoch() - 1.nanoseconds()).unix_timestamp()` is set to `0`, so it should be serialized as `timestamp 96`, but it is serialized as `timestamp 64`.

You can avoid this problem by changing `datetime >= OffsetDateTime::unix_epoch() && datetime.unix_timestamp() >> 34 == 0`, but there is still another problem.

If you serialize with `timestamp 96`, you have to separate nanoseconds from seconds, but if you use `unix_timestamp()` or `unix_timestamp_nanos() / 1_000_000_000`, the seconds will be `0`. It serializes `1970-01-01 0:00:00.99999999` instead of `1969-12-31 23:59:59.99999999`.

I would be very happy to solve the above problem with the development version v0.3 or v0.2 backported from v0.3.

Thank you for your consideration.

Please refer to the following
https://github.com/msgpack/msgpack/blob/master/spec.md#timestamp-extension-type

jhpratt:
I didn't take offense at all — just saying that in the future it's best to present a simple reason _why_ something is wrong, rather than just saying it is. Even in your most recent comment, you're assuming I have knowledge of the custom serialization format you're using, which is of course not the case.

Again, not an issue. Just some feedback for any future contributions to any project.

otake84:
Thanks for the merge.

You're right, I'll be careful from now on.

PR #304, opened 2021-01-11
Update dependencies, bump MSRV to 1.46

### LINKED ISSUE #298: Upgrading to rand 0.8

### LINKED ISSUE #302: Upgrade to quickcheck 1.0
The time crate should upgrade the quickcheck dependency to 1.0 prior to time's 0.3 release. The API is largely similar, but there are some changes. Notably, it is no longer possible (for the time, see BurntSushi/quickcheck#267) to generate a value within a given range. This can be worked around with the following function. It has some limitations with regard to overflow within the type, but I don't believe that should be an issue for any values used in the time crate.

```rust
fn arbitrary_between<T>(g: &mut Gen, min: T, max: T) -> T
where
    T: PartialOrd
        + AddAssign
        + Add<Output = T>
        + Sub<Output = T>
        + Rem<Output = T>
        + Arbitrary
        + Copy,
{
    #[allow(clippy::eq_op)]
    let zero = min - min;

    let range = max - min;
    let mut within_range = T::arbitrary(g) % range;

    if within_range < zero {
        within_range += range;
    }

    within_range + min
}
```

One thing I noticed in the documentation that I didn't before (I'm not sure if it was present or not and can't be bothered to check) is that the `Gen::size()` method _does not need to be used_, as all types in the time crate have a constant size; the example used as a use case is `Vec`, which most certainly does not. As such, the entire `foo_respects_generator_size` and `test_generator_size!` bit can be dropped, I believe.

This should likely be done in coordination with #299, though it's not strictly necessary. That change by itself will be nearly trivial.

cc @emlun if you're interested with your macro magic. No big deal if not.

jhpratt:
This upgrades quickcheck to 1.0 and rand to 0.8, in addition to bumping patch versions crate-wide.

From my reading of quickcheck's documentation, there is no reason to use `Gen::size` here, as that's intended for values without a statically known size (namely heap-allocated ones); that is not the case here. In response, such tests have been removed in their entirety.

Quickcheck now has `#![no_std]` support, though that's probably not particularly useful on its own.

This PR depends on BurntSushi/quickcheck#268 being resolved. CI is currently expected to fail, but should pass once a new release is cut (and this PR is updated to depend on it).

Closes #298, closes #302


jhpratt:
As quickcheck requires Rust 1.46, I'm going to bump the MSRV as part of this PR. This will allow a number of methods to always be `const fn` and the elimination of `proc-macro-hack` as a dependency when macros are enabled.

jhpratt:
Aside from the issue with quickcheck (which apparently was a bit more than I initially realized), this PR should be good to go.

cc @emlun; I know you expressed some interest in reviewing an upgrade. The first commit is the one that actually upgrades the dependencies; the second is just to make CI pass…eventually.

jhpratt:
Due to quickcheck 1.0 expanding the range of integers that is tested to the full range (as opposed to `-100..=100`), I'm getting a stack overflow when validating the ability to shrink the nanosecond portion of a `Duration` — this is because there's 2<sup>64</sup> possibilities for the valid seconds, so it's obviously not terminating. As such, attaining a portion of the iterator where the nanosecond value is changing takes effectively forever (far larger than the stack).

@emlun && @BurntSushi Any ideas on this? I may very well be overthinking this. I suppose it's not strictly necessary to validating the ability to shrink, though it is nice to have.

Unrelated, but when looking through quickcheck's implementations for `std::time::Duration`, I noticed the ability to shrink a tuple. So I'll be rewriting the code to do that, as it saves a lot of boilerplate.

BurntSushi:
I'm not quite sure I understand. Are you saying you're trying to iterate over every element yielded by `shrink`?

jhpratt:
The way the test works it's by ensuring that _some_ value in the iterator is closer to zero. It's being done by component, which was a non-issue with the limited range before. But now I'm fairly certain the shrinker recurses enough to overflow the stack. (side note: how on earth do you debug a stack overflow in Rust?)

This was originally implemented by Emil, so he'll understand the code better than me; my understanding has certainly grown, but it's far from complete.

BurntSushi:
I usually use gdb for stack overflows. I don't use it enough to have instructions precisely memorized, but basically, you run something like `gdb path/to/test/executable` and then in the gdb prompt, something like, `run {cli args}`. That should produce the stack overflow. Then run `bt` in the gdb prompt to get a full backtrace.

jhpratt:
Alright, I'll give that a shot when I'm on my laptop to confirm my theory at s minimum.

jhpratt:
Not entirely sure why, but switching the order of the nanoseconds and seconds in the tuple made things work fine with no additional effort needed.

Going to do a final check of the changes and merge.

emlun:
Sorry for being slow on this, but it looks pretty good to me! Delegating `shrink()` implementations down to shrinking tuples of shrinkable components is definitely the way to go, I think. Nice to see how much simpler that makes the code.

It looks to me like `arbitrary_between` might have ended up using `max` as an exclusive bound, instead of inclusive as the comments claim, but I'm not sure.

jhpratt:
I'll check that out later; it's certainly possible. I'll probably just wind up essentially brute forcing an 8 bit integer to verify it.

No worries on speed, by the way. Tis all volunteer regardless!

PR #313, opened 2021-03-01
Fix incorrect version constraint in Cargo.toml

jyn514:
`time` requires at least `version_check` 0.9.2 in order to use `to_mmp()`, but it says
in Cargo.toml that it can use 0.9.0. This broke docs.rs: https://github.com/rust-lang/docs.rs/pull/1295/checks?check_run_id=2007093638

This fixes the version constraint so Cargo will automatically update
version_check if necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ winapi = { version = "0.3", features = ["minwinbase", "minwindef", "timezoneapi"
 stdweb = { version = "0.4", default-features = false, optional = true }
 
 [build-dependencies]
-version_check = "0.9"
+version_check = "0.9.2"
 
 [dev-dependencies]
 rand = { version = "0.7", default-features = false }


jhpratt:
Is there any reason running `cargo update` does not suffice? That's the reason I haven't made this change previously — even as recently as this past week in #311. I currently have no plans to publish another release in the 0.2 series, as 0.3 is nearly complete.

jyn514:
> Is there any reason running cargo update does not suffice?

`cargo update` tends to lead to production outages: https://github.com/rust-lang/docs.rs/pull/1278. I ran `cargo update -p version_check` in the meantime which worked fine, but I thought it would be nice to avoid this in the future.

jhpratt:
That's…odd. The main branch uses the most recent releases (as of whenever I last upgraded them, at least), so this issue won't arise in 0.3.

jyn514:
> That's .... odd.

Do you mean that it caused an outage? That's a whole separate thing, it doesn't have anything to do with time.

Anyway, I don't particularly care whether this lands or not now that I've fixed in on our end, but you may want it just so you  stop getting bug reports ;)

jhpratt:
> Do you mean that it caused an outage? That's a whole separate thing, it doesn't have anything to do with time.

Gotcha. I didn't check the link at first, clearly.

As I stated in the recent PR, 0.3 is nearing release and no further 0.2-series releases are planned, so going ahead and closing this.

PR #316, opened 2021-03-08
[v0.2] Fix feature detection with third-party rustc

cuviper:
Third-party `rustc` builds, like those in Linux distributions, are built
from the source tarball without git commit info in `rustc --version`.
This makes `version_check::triple()` fail, which makes `time` unhappy:

    warning: Unable to determine rustc version. Assuming rustc 1.32.0.

The build script doesn't actually care about the commit date though, and
it works fine if you only request the version and channel separately.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -26,11 +26,12 @@ fn main() {
 
     // Treat nightly & dev compilers as the equivalent of the then-beta.
     // As features are never stabilized in patch versions, we can safely ignore it.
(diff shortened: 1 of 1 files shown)


jhpratt:
I'll get around to this early next week. While technically not a bug, as everything still works fine, I'm going to put out a patch release as if this were a bug fix. It'll give me a chance to set proper patch versions of the dependencies as well, which has been brought up a couple times.

jhpratt (approved):

PR #319, opened 2021-03-24
Update whatrustisit URL

atouchet:
I get a security warning when clicking this badge. Re-adding the www fixes the issue.


jhpratt:
imo this should be filed as an issue in the [badboy/whatrustisit](https://github.com/badboy/whatrustisit) repository, which appears to be the source for the website. It's a configuration issue with the site, after all.

PR #320, opened 2021-03-27
fix documentation for %C

soenkehahn:
As stated [here](https://github.com/time-rs/time/blob/v0.2/src/format/date.rs#L138), the `%C` format specifier accepts both 2 and 3 digits.


jhpratt:
CI failure is unrelated. As another release for 0.2 is not planned, this likely won't ever end up on docs.rs, but I'll merge this regardless.

PR #326, opened 2021-05-17
Remove WINAPI dependency, replace with FFI calls

Adamcake:
The dependency on the `winapi` crate here, which is over 10MB and has some considerable impact on build and link times, is quite unnecessary - it's only used for one function in one file, namely `UtcOffset::local_offset_at()` when the `local-offset` feature is enabled. This PR removes the crate as a dependency and replaces its usage in that function with a few lines of FFI.

The function still returns the same results, but with a significantly better build time. All tests pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ default = ["std"]
 alloc = []
 formatting = ["itoa", "std"]
 large-dates = ["time-macros/large-dates"] # use case for weak feature dependencies (rust-lang/cargo#8832)
-local-offset = ["std", "winapi"]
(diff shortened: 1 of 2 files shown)


jhpratt:
Does Windows guarantee the order of the various fields? I recall that being an issue with Unix-like systems, which is why libc is still used.

jhpratt:
Merged as I couldn't find a way to run CI. I manually checked that it type-checks (I develop on Linux, so couldn't actually test). I'm fairly certain CI will pass on the merged commit, but I'll revert if it doesn't.

PR #346, opened 2021-07-31
Add integer From/TryFrom impls for Month

### LINKED ISSUE #345: Month conversion to/from integral numbers
**Preamble**

As of version 0.3.0, the `Month` enum is a complete black box. It has no safe way of converting to/from numbers, and it even intentionally doesn't implement `Ord` or `PartialOrd`.

In my experience, there are valid use cases where you want those features. Here's one example:

```rust
pub fn age_at(date: Date, birthday: Date) -> Year {
    let (year, month, day) = date.to_calendar_date();
    let (year_b, month_b, day_b) = birthday.to_calendar_date();

    if (month_b.ordinal(), day_b) <= (month.ordinal(), day) {
        year - year_b
    } else {
        year - year_b - 1
    }
}
```

Another use case is date arithmetic. I realize that "date + 10 months" is a giant can of worms, but sometimes you have no choice but to crack that can open.

**Proposal**

How about `impl From<Month> for u8` and `impl TryFrom<u8> for Month`? I feel like these are fairly reasonable things to add to the library. The error type for the latter would be `time::error::ComponentRange`.

dmit:
Closes #345.

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -128,17 +128,15 @@ impl TryFrom<u8> for Month {
     type Error = error::ComponentRange;
 
     fn try_from(value: u8) -> Result<Self, Self::Error> {
-        NonZeroU8::new(value).map_or_else(
-            || {
-                Err(error::ComponentRange {
-                    name: "month",
-                    minimum: 1,
-                    maximum: 12,
-                    value: value as _,
-                    conditional_range: false,
-                })
-            },
-            Self::from_number,
-        )
+        match NonZeroU8::new(value) {
+            Some(value) => Self::from_number(value),
+            None => Err(error::ComponentRange {
+                name: "month",
+                minimum: 1,
+                maximum: 12,
+                value: 0,
+                conditional_range: false,
+            }),
+        }
     }
 }
--- a/src/month.rs
+++ b/src/month.rs
@@ -128,13 +128,17 @@ impl TryFrom<u8> for Month {
     type Error = error::ComponentRange;
 
     fn try_from(value: u8) -> Result<Self, Self::Error> {
-        let n = NonZeroU8::new(value).ok_or_else(|| error::ComponentRange {
-            name: "month",
-            minimum: 1,
-            maximum: 12,
-            value: value.into(),
-            conditional_range: false,
-        })?;
-        Self::from_number(n)
+        NonZeroU8::new(value).map_or_else(
+            || {
+                Err(error::ComponentRange {
+                    name: "month",
+                    minimum: 1,
+                    maximum: 12,
+                    value: value as _,
+                    conditional_range: false,
+                })
+            },
+            Self::from_number,
+        )
     }
 }
--- a/src/month.rs
+++ b/src/month.rs
@@ -1,12 +1,11 @@
 //! The `Month` enum and its associated `impl`s.
 
+use core::convert::TryFrom;
 use core::fmt;
-#[cfg(feature = "parsing")]
 use core::num::NonZeroU8;
 
 use self::Month::*;
-#[cfg(feature = "parsing")]
-use crate::error;
+use crate::error::ComponentRange;
 
 /// Months of the year.
 #[allow(clippy::missing_docs_in_private_items)] // variants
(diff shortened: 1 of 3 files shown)


jhpratt:
Forgot to update the required status checks when I bumped the MSRV. I'll do that in a bit.

src/month.rs
```diff
@@ -1,12 +1,11 @@
...
 
+use core::convert::TryFrom;
 use core::fmt;
-#[cfg(feature = "parsing")]
 use core::num::NonZeroU8;
 
 use self::Month::*;
-#[cfg(feature = "parsing")]
-use crate::error;
+use crate::error::ComponentRange;
```
jhpratt, line 8: Let's keep the `use crate::error` import and refer to it as `error::ComponentRange`. Without this, it's not immediately clear that the type is an error type.

src/month.rs
```diff
@@ -119,3 +117,37 @@ impl fmt::Display for Month {
...
+            April => 4,
+            May => 5,
+            June => 6,
+            July => 7,
+            August => 8,
+            September => 9,
+            October => 10,
+            November => 11,
+            December => 12,
+        }
```
jhpratt, line 136: This could just be `month as _`

tests/integration/month.rs
```diff
@@ -1,4 +1,6 @@
-use time::Month::*;
+use std::convert::TryFrom;
+
+use time::{Month, Month::*};
```
jhpratt, line 3:
```suggestion
use time::Month::{self, *};
```

Disregard if rustfmt doesn't like this.

src/month.rs
```diff
@@ -119,3 +117,37 @@ impl fmt::Display for Month {
...
+    }
+}
+
+impl TryFrom<u8> for Month {
+    type Error = ComponentRange;
+
+    fn try_from(value: u8) -> Result<Self, ComponentRange> {
```
jhpratt, line 143:
```suggestion
    fn try_from(value: u8) -> Result<Self, Self::Error> {
```

src/month.rs
```diff
@@ -119,3 +117,37 @@ impl fmt::Display for Month {
...
+        let n = NonZeroU8::new(value).ok_or_else(|| ComponentRange {
+            name: "month",
+            minimum: 1,
+            maximum: 12,
+            value: value.into(),
+            conditional_range: false,
+        })?;
+        Self::from_number(n)
```
jhpratt, line 151:
```suggestion
        NonZeroU8::new(value).map_or_else(Self::from_number, || Err(ComponentRange {
            name: "month",
            minimum: 1,
            maximum: 12,
            value: value as _,
            conditional_range: false,
        }))
```

I haven't checked this, but I'm pretty sure it'll compile.

tests/integration/month.rs
```diff
@@ -47,3 +49,37 @@ fn display() {
...
+    assert_eq!(u8::from(August), 8u8);
+    assert_eq!(u8::from(September), 9u8);
+    assert_eq!(u8::from(October), 10u8);
+    assert_eq!(u8::from(November), 11u8);
+    assert_eq!(u8::from(December), 12u8);
+}
+
+#[test]
+fn try_from_u8() {
+    assert!(Month::try_from(0u8).is_err());
```
jhpratt, line 71: It would be good to assert the component name here.

```suggestion
    assert!(matches!(Month::try_from(0u8), Err(err) if err.name() == "month"));
```

jhpratt (changes requested):

src/month.rs
dmit: This one fails to compile because the first closure in `map_or_else` takes no arguments, but `from_number` takes an `NonZeroU8`.

jhpratt: Ah, right. Flip the order of the two arguments. I forgot the "else" bit comes first.

dmit:
All suggestions addressed. Please let me know if you want me to squash the PR into a single commit.

jhpratt:
Actually just thought of the fact that the only time the error part of `map_or_else` is when the value is zero. I can handle that myself.

Right now I'm trying to figure out how to run the CI. Apparently it's disabled by default for first-time contributors, but I don't see a way to explicitly run it.

jhpratt (approved):

jhpratt:
No idea how to get CI to run, since the button that GitHub docs say should be there isn't. Regardless I've tested this locally in a few ways and am confident enough to merge it.

PR #355, opened 2021-10-02
Lowercase lib name in Windows link attribute

Skirmisher:
Cross-building for Windows on (for example) Linux exposes some fun case-sensitivity issues, since mingw libraries are in all lowercase. This just changes title-case library names to lowercase, and shouldn't affect native Windows builds at all (I hope).


jhpratt:
Is there a setup that could be used in CI to ensure this doesn't regress?

CI isn't running because you're a first-time contributor here, but I _should_ have a button that runs it manually. For whatever reason that isn't the case, which is why CI appears as though it had stalled.

Skirmisher:
I have approximately zero experience with Github Actions... but, it *should* be pretty straightforward to install the [mingw-w64](https://packages.ubuntu.com/focal/mingw-w64) toolchain on Ubuntu and run something like `cargo build --tests --target x86_64-pc-windows-gnu` (might need some `RUSTFLAGS` to pass the mingw library path, not sure). (Though I haven't been able to build the tests myself for some reason...)

jhpratt:
For status I'm just waiting on a response from GitHub on why I can't run CI here. I can add a regression check in CI myself.

jhpratt:
After doing that and applying this patch, I still get an error.

```text
   Compiling time v0.3.3 (/home/jhpratt/code/time)
error: linking with `x86_64-w64-mingw32-gcc` failed: exit status: 1
  |
  = note: "x86_64-w64-mingw32-gcc" <omitted>
  = note: /usr/lib/gcc/x86_64-w64-mingw32/10.3.1/../../../../x86_64-w64-mingw32/bin/ld: cannot find -l:libpthread.a
          collect2: error: ld returned 1 exit status
```

PR #363, opened 2021-10-13
Make `T::default` for format_description types public API, and mention the default values of the types in their docs.

### LINKED ISSUE #362: Add `const fn new() -> Self` to `time::format_description` types?
I know you have the `#[doc(hidden)] const fn default() -> Self` so that things like `time::format_description::modifier::Year::default()` work, and I see the comment that it's a hack until `Default::default()` by itself becomes legal in a const context. But since there are no plans for doing that (RFC 2632 was closed in 2021-07), it would be nice to have a non-`#[doc(hidden)]` of doing that, like a documented `const fn new() -> Self`.

I was actually going to make a PR to add a `const fn new() -> Self` to the macro output when I discovered this existing `#[doc(hidden)]` function. I imagine there must be other users who came in with the prior knowledge of `Default::default()` not working, didn't see any `const` constructors in the documentation, and assumed it just isn't possible to construct these values in a const context.

(Also, while it's true that an `fn new() -> Self` without parameters is redundant with a `Default` impl in general, it's still better for discoverability to have an `fn new`.)

Arnavion:
Fixes #362

### DIFF
--- a/src/format_description/modifier.rs
+++ b/src/format_description/modifier.rs
@@ -242,22 +242,40 @@ pub enum Padding {
     None,
 }
 
+macro_rules! doc_hack {
+    (pub $(#[$attr:meta])*; $($x:tt)*) => {
+        $(#[$attr])*
+        ///
+        /// This function exists since [`Default::default()`] cannot be used in a `const` context.
+        /// It may be removed once that becomes possible. As the [`Default`] trait is in the
+        /// prelude, removing this function in the future will not cause any resolution failures for
+        /// the overwhelming majority of users; only users who use `#![no_implicit_prelude]` will be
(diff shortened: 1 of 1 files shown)


jhpratt:
After going through this locally, I've made a few changes that aren't notable (mostly just wording and formatting). However, I did notice one thing new. This method is being added for _all_ types in the module, not just those where it's necessary. `#[non_exhaustive]` structs need this hack but enums do not — they can be constructed already as long as the desired variant is exhaustive. Given that the relevant enums are all C-like, there's no need to include them when exposing this method.

I can manage that change myself through some macro trickery. I'm going to do that, sleep on it, and merge.

PR #368, opened 2021-10-20
refactor: parse_* methods accept more argument types

### LINKED ISSUE #365: Date::parse now doesn't accept String
(Other `parse` methods as well, I think)

While it used to accept it and now doesn't, there's no real reason for it other than some method signatures changing during the big rewrite - so it needlessly breaks some code that used to work in 0.2.

The solution is simple: take `impl AsRef<str>` instead of `&str`.

seanpianka:
By using AsRef<str>, the `parse_*` methods can accept arguments of different types as long as they can be converted to a `&str`.

Closes #365

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -685,11 +685,11 @@ impl Date {
     /// assert_eq!(Date::parse("2020-01-02", &format)?, date!(2020 - 01 - 02));
     /// # Ok::<_, time::Error>(())
     /// ```
-    pub fn parse(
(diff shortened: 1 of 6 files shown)


jhpratt:
In addition to the note already left, we don't actually need a `&str`, as parsing operates entirely on bytes. The conversion is made a few lines down where `.as_bytes()` is called. However, I am still unconvinced that this change needs to be made at all.

Issue #339, opened 2021-07-26
Microsoft JSON dates


fossdd:
I just come on a error on parsing with `time` that it doesn't regonized the type of microsoft's JSON used date.

A example is:
```
/Date(151708196000)/
```

The `/Date(` and the `)/` can be removed, and then there is a known unix timestamp. Would nice if `time` could implement it.

In that time, how am I now able to parse this format?

jhpratt:
Currently you have to create customer serializer/deserializer functions to do this. By default serializing and deserializing use a single format — it will never try another if the first fails. In the future I hope to have something like `#[serde(with = time::serde::format_description!("/Date([timestamp])/")]` which would auto-generate a module.

jhpratt:
The `[unix_timestamp]` component is now implemented and documented alongside the other components.

PR #381, opened 2021-11-02
Add macros for custom serde formats

znewman01:
Hi!  This makes progress towards (but does not fix) #339 (which still needs a `[timestamp]` format component). Regardless, I would use it in its current implementation.

This is what it looks like:

```rust
declare_format_string!(my_format, "hour=[hour], minute=[minute]");

#[derive(Serialize, Deserialize)]
struct SerializesWithCustom {
    #[serde(with = "my_format")]
    dt: OffsetDateTime,
}
```

Is this something you'd be interested in merging in (roughly) its current shape? If so, I'm happy to take care of renaming/documenting things.

## Design notes

- I couldn't figure out a way to make `#[serde(with = my_macro!())]` work, since `with` expects a string.
- Similarly, wasn't sure whether ` - names were the first thing that popped into my head; please give better suggestions :)
- I may not have handled the feature-gating correctly.
- Clippy is complaining about `to_invalid_serde_value` because `error::Format` [isn't copy][clippy], so I had to call it `into_invalid_serde_value`.
- If we want to implement these serializer modules in other crates (a normal use case), ,we have to make the `{in,}to_invalid_serde_value` functions public

[clippy]: https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention


jhpratt (changes requested):
Initial thoughts inline. Additionally CI must pass and coverage of the patch should be 100%. The macro name is really the least important thing as that can be trivially changed.

src/error/component_range.rs
```diff
@@ -77,7 +77,7 @@ impl serde::de::Expected for ComponentRange {
 #[cfg(feature = "serde")]
 impl ComponentRange {
     /// Obtain an "invalid value" error type for the deserializer.
-    pub(crate) fn to_invalid_serde_value<'a, D: serde::Deserializer<'a>>(self) -> D::Error {
+    pub fn to_invalid_serde_value<'a, D: serde::Deserializer<'a>>(self) -> D::Error {
```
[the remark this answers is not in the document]
znewman01: If we want users to be able to perform this conversion from outside of the crate (even inside a macro), we need to expose it.

As far as I can see our options are:
1. make them `pub`
2. reimplement inside of the macro each time

Unless I'm missing a better option. I'm happy to do (2) if you'd prefer.

znewman01:
> Initial thoughts inline. 

Appreciate the prompt feedback; I can address everything you mentioned but would appreciate guidance re:exposing  `to_invalid_serde_value` first; after that, should be able to take another pass in a couple days.

> Additionally CI must pass and coverage of the patch should be 100%.

Of course, will take care of that as well (wanted to make sure you'd actually be interested in the PR first).

src/error/component_range.rs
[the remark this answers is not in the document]
jhpratt: Yeah, it's unfortunate that Rust doesn't have the ability for workspace-public visibility, though I certainly understand the reasoning (it's more difficult and has more blockers than you'd expect). I suppose this could be exposed and be `#[doc(hidden)]` similar to a few other methods that are only `pub` for macros. In those situations they're `pub` for additional reasons, admittedly.

jhpratt:
I'll give this a proper look early next week most likely. Until then I just checked the CI logs. You should be able to replace `format!` with `format_args!`, which will work in no-alloc situations and probably be a little more efficient anyways. For coverage you'll need to run `TRYBUILD=overwrite RUSTFLAGS="--cfg __ui_tests" cargo test --all-features --workspace` and commit the resulting files.

znewman01:
Feel free to wait another couple days---I didn't have time to wrap this up. I should definitely get to it this week.

jhpratt:
Not an issue. Let me know when it's ready.

src/error/component_range.rs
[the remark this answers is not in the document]
znewman01: I went with `#[doc(hidden)]`

src/error/format.rs
```diff
@@ -85,3 +85,21 @@ impl TryFrom<crate::Error> for Format {
...
 }
+
+#[cfg(feature = "serde")]
+impl Format {
+    /// Obtain an error type for the serializer.
+    pub fn into_invalid_serde_value<S: serde::Serializer>(self) -> S::Error {
+        use serde::ser::Error;
+        match self {
+            Self::InsufficientTypeInformation => {
+                S::Error::custom("Insufficient type information to format a component.")
```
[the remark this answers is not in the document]
znewman01: Were you referring to the convention on capitalization/punctuation? If so, done.

src/serde/macros.rs
```diff
@@ -0,0 +1,126 @@
...
+/// [`FormatItem`]: crate::format_description::FormatItem
+/// [`Rfc3339`]: crate::format_description::well_known::Rfc3339
+#[macro_export]
+macro_rules! declare_format {
+    ($i:ident, $e:tt) => {
+        mod $i {
+            time::declare_format!($e);
+        }
+    };
+    ($e:tt) => {
```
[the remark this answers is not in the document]
znewman01: Good point -- removed that option.

src/serde/macros.rs
```diff
@@ -0,0 +1,126 @@
+//! Provides convenient way for declaring custom serializer/deserializer modules.
+//!
+//! Use this module in combination with serde's [`#[with]`][with] attribute.
+//!
+//!
+//! [with]: https://serde.rs/field-attrs.html#with
+
+/// Declares a custom format based on the given [`FormatItem`].
```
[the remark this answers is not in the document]
znewman01: I believe the doc string is better now.

tests/integration/serde_macros.rs
```diff
@@ -0,0 +1,116 @@
...
+    };
+    assert_tokens(
+        &value.compact(),
+        &[
+            Token::Struct {
+                name: "TestCustomFormat",
+                len: 1,
+            },
+            Token::Str("dt"),
+            Token::BorrowedStr("custom format: 2000-01-01 00:00:00 -04:00"),
```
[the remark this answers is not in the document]
znewman01: Sorry, not sure what you're getting at here. The custom format I have includes `[second]` which seems to unconditionally include the seconds value.

I also tried adding a deserialize test for `custom format: 2000-01-01 00:00: -04:00` (omitting seconds), and I get an error: `invalid value: second, expected valid second`.

znewman01:
@jhpratt I think it's ready for another pass.

Sorry it took so long, I ran into a few snags.

I needed a serde serializer to `Rfc3339` to cover the `InvalidComponent` branch of `error::Format::into_invalid_serde_value` so I wrote one before noticing the `serde-rfc3339` branch. I'm happy to rebase on that and use your implementation instead (though I noticed it just uses serde's `Error::custom` for everything--should I prefer that here, too?).

The other big change is that I needed the `PrimitiveDateTime` analog of this macro for coverage. Should I keep going with `Time`, `Date`, and `UtcOffset`?

jhpratt:
Don't worry about the time; I've got other things going on and I'm sure you do as well. It's not like there's a huge rush to put this out.

> I needed a serde serializer to `Rfc3339` to cover the `InvalidComponent` branch of `error::Format::into_invalid_serde_value` so I wrote one before noticing the `serde-rfc3339` branch. I'm happy to rebase on that and use your implementation instead (though I noticed it just uses serde's `Error::custom` for everything--should I prefer that here, too?).

Definitely don't rebase on that branch, as it's blocked on an upstream change in Cargo. That said, feel free to pull something specific if you need it. The only requirement is that enabling macros shouldn't enable serde.

For `Error::custom`, it doesn't much matter. Internally serde just calls that anyways. So if you think it's simpler/easier, go for it.

> The other big change is that I needed the `PrimitiveDateTime` analog of this macro for coverage. Should I keep going with `Time`, `Date`, and `UtcOffset`?

The macro could accept a type name. Something like `declare_format_string!(Date; "[year]-[month]-[day]");`

tests/integration/serde/timestamps.rs
```diff
@@ -0,0 +1,42 @@
+use serde::{Deserialize, Serialize};
+use serde_test::{assert_de_tokens_error, assert_tokens, Configure, Token};
+use time::macros::datetime;
+use time::serde::timestamp;
+use time::OffsetDateTime;
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
```
jhpratt, line 7: I don't think we actually need the derives other than serde's.

time-macros/src/error.rs
```diff
@@ -114,4 +114,30 @@ impl Error {
...
+            TokenTree::from(Ident::new("compile_error", start)),
+            TokenTree::from(Punct::new('!', Spacing::Alone)).with_span(start),
+            TokenTree::from(Group::new(
+                Delimiter::Parenthesis,
+                TokenStream::from(
+                    TokenTree::from(Literal::string(&self.to_string())).with_span(end),
+                ),
+            ))
+            .with_span(start),
+            TokenTree::from(Punct::new(';', Spacing::Alone)).with_span(end),
```
jhpratt, line 137: The only difference is this semicolon, correct? Is there any reason we can't call `to_compile_error()` and then just append this single token?

time-macros/src/lib.rs
```diff
@@ -89,3 +91,114 @@ pub fn format_description(input: TokenStream) -> TokenStream {
...
 }
+
+fn make_serde_serializer_module(
+    mod_name: proc_macro::Ident,
+    items: impl to_tokens::ToTokens,
+    formattable: TokenStream,
+) -> TokenStream {
+    let serialize_fns = quote! {
+        pub fn serialize<S: ::serde::Serializer>(
+                datetime: &#(formattable.clone()),
```
jhpratt, line 102: Check the formatting on code inside `quote!` — rustfmt doesn't handle it.

time-macros/src/lib.rs
```diff
@@ -89,3 +91,114 @@ pub fn format_description(input: TokenStream) -> TokenStream {
...
+        Some(tree) => {
+            return Error::UnexpectedToken { tree }.to_compile_error_standalone();
+        }
+        None => return Error::UnexpectedEndOfInput.to_compile_error_standalone(),
+    };
+    // Followed by a comma
+    match tokens.next() {
+        Some(tree) => {
+            if let TokenTree::Punct(ref punct) = tree {
+                if punct.as_char() != ',' {
```
jhpratt, line 171:
```suggestion
            if matches!(tree, TokenTree::Punct(ref punct) if punct.as_char() != ',') {
```

Also might be able to omit the `.as_char()`. I forget if it was stabilized before or after the MSRV.

src/error/parse.rs
```diff
@@ -104,13 +104,34 @@ impl TryFrom<crate::Error> for Parse {
...
+///
+/// This allows us to use `format_args!()` in a `serde::de::Error`.
+#[cfg(feature = "serde")]
+struct ArgumentsWrapper<'a> {
+    /// Inner wrapped value.
+    args: fmt::Arguments<'a>,
+}
+
+#[cfg(feature = "serde")]
+impl<'a> From<fmt::Arguments<'a>> for ArgumentsWrapper<'a> {
```
jhpratt, line 117: These lifetimes can be elided.

jhpratt (changes requested):

tests/integration/compile-fail/invalid_serializer.rs
```diff
@@ -0,0 +1,21 @@
+use time::macros::{
+    declare_format_string_offset_date_time, declare_format_string_primitive_date_time,
+};
+
+declare_format_string_offset_date_time!(); // unexpected end of input
```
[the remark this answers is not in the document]
znewman01: Did you mean something other than `invalid_serializer.stderr`? I'm not sure what you're asking for.

(The comments were just there so I can keep track of the possible failure methods while looking at the test; if you'd like, I can remove them.)

tests/integration/serde/timestamps.rs
znewman01: I can remove `Eq` but everything else is required by [`assert_tokens`](https://docs.rs/serde_test/1.0.130/serde_test/fn.assert_tokens.html)

time-macros/src/lib.rs
znewman01: Looks like the [`PartialEq<char> for Punct`](https://doc.rust-lang.org/stable/proc_macro/struct.Punct.html#impl-PartialEq%3Cchar%3E) was added in 1.50 so we're good

time-macros/src/lib.rs
znewman01: I think I got it now

src/error/parse.rs
znewman01: I'm not sure how -- we need to specify that the `ArgumentsWrapper` lives no longer than the `fmt::Arguments`.

I couldn't come up with any combination of elisions that rustc was happy with.

tests/integration/serde/macros.rs
```diff
@@ -0,0 +1,149 @@
...
+            Token::BorrowedStr("custom format: 2000-01-01 00:00:00"),
+            Token::Str("primitive_option"),
+            Token::Some,
+            Token::BorrowedStr("custom format: 2000-01-01 00:00:00"),
+            Token::StructEnd,
+        ],
+    );
+    let value = TestCustomFormat {
+        offset_dt: datetime!(2000-01-01 00:00 -4:00),
+        offset_option: None,
```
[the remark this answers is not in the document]
znewman01: Good thought, done.

znewman01:
Okay, I've addressed most of the specific comments (would appreciate feedback on the ones still open). I have three higher-level changes to make, too:

- [ ] Just use `Error::custom` for `error::Format` serde conversion
- [ ] Get rid of my `Rfc3339` implementation, which is no longer necessary for coverage, in favor of the existing branch
- [ ] Pass the type name to the macro

tests/integration/serde/mod.rs
```diff
@@ -2,6 +2,12 @@ use serde_test::{assert_de_tokens_error, assert_tokens, Compact, Configure, Read
 use time::macros::{date, datetime, offset, time};
 use time::{Date, Duration, Month, OffsetDateTime, PrimitiveDateTime, Time, UtcOffset, Weekday};
 
+mod error_conditions;
+#[cfg(feature = "macros")]
```
jhpratt, line 6: Tests are always run with `--all-features`, so this isn't strictly necessary.

jhpratt (commented):
For the RFC3339 implementation, feel free to keep it if it doesn't require additional features like the existing branch does. That's the only reason I haven't merged that branch in.

For `Error::custom`, I think it should just be a matter of replacing the existing calls. If there's a specific problem you're having here, let me know and I'll see what I can do.

For passing the type name to the macro, right now you can just assume it'll be the identifier alone and not a path. It's possible to preserve the hygiene such that it's on the caller to have it in scope. I can handle accepting a path rather than just an identifier after merging, as that'll require an additional helper that's not trivial.

tests/integration/serde_macros.rs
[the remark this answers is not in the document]
jhpratt: Whoops, you're right. The only place my comment applies is in macro invocations.

Cargo.toml
```diff
@@ -33,7 +33,7 @@ local-offset = ["std"]
 macros = ["time-macros"]
 parsing = []
 quickcheck = ["quickcheck-dep", "alloc"]
-serde-human-readable = ["serde", "formatting", "parsing"]
+serde-human-readable = ["serde/alloc", "formatting", "parsing"]
```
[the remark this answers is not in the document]
znewman01: > Is it the RFC3339 format that requires this change? 

Correct. I think I'll keep it like this for now.

src/error/parse.rs
znewman01: No, it doesn't in this case.

znewman01:
@jhpratt ready for another look!

> For the RFC3339 implementation, feel free to keep it if it doesn't require additional features like the existing branch does. That's the only reason I haven't merged that branch in.

Okay, great!

> For Error::custom, I think it should just be a matter of replacing the existing calls. If there's a specific problem you're having here, let me know and I'll see what I can do.

Well, if we're leaving RFC3339 in I'll keep the errors as they are because I can get full coverage now.

> For passing the type name to the macro, right now you can just assume it'll be the identifier alone and not a path. It's possible to preserve the hygiene such that it's on the caller to have it in scope. I can handle accepting a path rather than just an identifier after merging, as that'll require an additional helper that's not trivial.

Much easier, done.

jhpratt:
I'm going to get around to this at some point in the near future. In part due to the size of the PR, I'm prioritizing other things first. I also want to cut a release with a few patches before finalizing my decision on the exact manner this will take, so having other things land first is strategic.

time-macros/src/lib.rs
```diff
@@ -89,3 +91,109 @@ pub fn format_description(input: TokenStream) -> TokenStream {
...
+            Option::<&str>::deserialize(deserializer)?
+                .map(|string| #(formattable.clone())::parse(string, &DESCRIPTION))
+                .transpose()
+                .map_err(::time::error::Parse::to_invalid_serde_value::<D>)
+        }
+    };
+
+    quote! {
+        mod #(mod_name) {
+            use ::time::#(formattable.clone());
```
jhpratt, line 145: No need to change this; I'm adding this as a note for myself to think about.

Do we want the implicit import or do we want to make the user specify the path if it's not in scope? The parser would need to be slightly adjusted to accept a path instead of an identifier.

time-macros/src/lib.rs
```diff
@@ -89,3 +91,109 @@ pub fn format_description(input: TokenStream) -> TokenStream {
...
+        }
+        None => return Error::UnexpectedEndOfInput.to_compile_error_standalone(),
+    };
+    // Followed by a comma
+    if let Err(err) = helpers::consume_punct(',', &mut tokens) {
+        return err.to_compile_error_standalone();
+    }
+    // Then, the type to create serde serializers for (e.g., `OffsetDateTime`).
+    let formattable = match tokens.next() {
+        Some(TokenTree::Ident(ident)) => TokenTree::Ident(ident),
```
jhpratt, line 177: This should be able to be replaced with `Some(tree @ TokenTree::Ident(_)) => tree`, which avoids creating a new value.

jhpratt:
Looks good overall. Mostly just waiting on me at this point.

jhpratt:
I've actually changed my mind on RFC3339. I believe I'll be able to get the existing branch working without much effort, and it's not inherently related to the macros for serde, so I think it would be best for it to be removed.

Sorry for the flip-flop.

znewman01:
I should be able to take a pass in the next couple days and address the above.

> I've actually changed my mind on RFC3339. I believe I'll be able to get the existing branch working without much effort, and it's not inherently related to the macros for serde, so I think it would be best for it to be removed.

Sure, can do. To be clear, the plan is to rebase this on top of RFC3339 (we need that for full coverage) and then it (this branch) can go in after RFC3339? There's no urgency on my end. Do you have a sense of when the RFC3339 branch will go in?

tests/integration/compile-fail/invalid_serializer.rs
```diff
@@ -0,0 +1,17 @@
...
+declare_format_string!(my_format,); // missing formattable and string
+declare_format_string!(my_format, "[year] [month]"); // missing formattable
+declare_format_string!(OffsetDateTime, "[year] [month]"); // missing ident
+declare_format_string!(my_format, OffsetDateTime); // missing string format
+declare_format_string!(my_format, OffsetDateTime,); // missing string format
+declare_format_string!(my_format, OffsetDateTime "[year] [month]"); // missing comma
+declare_format_string!(my_format, OffsetDateTime : "[year] [month]"); // not a comma
+declare_format_string!(my_format, OffsetDateTime, "[bad]"); // bad component name
+declare_format_string!(my_format, OffsetDateTime, not_string); // string format wrong type
+declare_format_string!(my_format, Date, "[year] [month]"); // formattable not in scope
```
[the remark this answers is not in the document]
znewman01: Left-over from before, removed now

time-macros/src/lib.rs
znewman01: Nice, that's a new trick for me :)

znewman01:
Ok, I have rebased on `serde-rfc3339` and addressed all comments. I'm okay to wait until that's merged for this to go in.

If it's better to merge this PR first, I can remove the final commit and rebase on `main`. This will cause a decrease in coverage temporarily (maybe you can cherry-pick [7f3e192](https://github.com/time-rs/time/pull/381/commits/7f3e192b9717805699510d510cd7840ad7f4b6ba) on the `serde-rfc3339` branch).

Either way, I'm happy to rebase---the commit history has gotten a little unwieldy.

jhpratt:
Don't worry about the coverage — I see that the diff coverage is 100% which is what matters.

I'm going to work on getting RFC3339 (and now RFC2822) merged, then I'll take a look at this for what is hopefully the final time.

jhpratt:
Now that support for RFC3339 and RFC2822 in serde are on main, would you mind rebasing and removing that bit from this PR? Other than that this PR is probably fine. The commits can be squashed together, so the actual history doesn't much matter.

znewman01:
Coverage issue is because this adds back `error::Parse::to_invalid_serde_value` which was removed in [64f2ee](https://github.com/time-rs/time/commit/64f2ee56d1ede827987e1299485998b4ff44b1ce).

I can try to mimic that style and remove `to_invalid_serde_value` but will not get to it today.

jhpratt:
No worries. I'm taking a mini-break from things for a couple days anyways.

znewman01:
This latest push passes all checks, including coverage. Ready for your review.

znewman01:
@jhpratt ping (no pressure/urgency! just making sure this hasn't slipped off your radar)

jhpratt:
Still on my radar! I'm working on an RFC right now, hence the delay.

znewman01:
@jhpratt ping again, but please take it in the same spirit as above! (no pressure)

jhpratt:
No worries. I assure you I haven't forgotten about this :slightly_smiling_face: Just yesterday I started reviewing (and modifying as necessary) the open PRs in this repo.

znewman01:
Awesome :) thanks for your hard work maintaining this project!

jhpratt:
I've just rebased before merging. There's some changes I'd like to make, but they're relatively minor compared to the work you've already done. I think merging this now is beneficial because it'll force me to actually work on things as opposed to leaving this lying around for however much longer. Plus it ensures that merge conflicts are my problem, not yours.

jhpratt (approved):

Issue #384, opened 2021-11-03
`PrimitiveDateTime::assume_offset` may overflow


L117:
`assume_offset` calls `offset_to_utc`, which does not perform checks.
https://github.com/time-rs/time/blob/be29bb35f89524a2c9cd3738fe6dce1c5644e59e/src/primitive_date_time.rs#L531-L553
Sample code to reproduce:
```rust
use time::macros::{datetime, offset};

fn main() {
    let datetime = datetime!(-999999 - 01 - 01 00:00);
    let datetime = datetime
        .assume_offset(offset!(+10:00))
        .replace_offset(offset!(+00:00));
    assert_eq!(datetime.year(), -1_000_000);
}
```
In particular, this means methods `OffsetDateTime::checked_{add,sub}` contain some unexpected behavior when date value is out of range `Date::MIN ..= Date::MAX`. And their doctests also have overflowed values (Shame on me for not noticing this earlier).

I can write a PR, but what exactly should it change?

jhpratt:
Hey, for once I actually noticed this issue previously! See #307 for a detailed description and possible solutions. I previously decided the status quo was acceptable. I'm open to thoughts on other possible solutions (including drawbacks), given that I doubt many people saw #307 at the time.

L117:
Well, personally I'd always choose correctness over convenience. Luckily, it seems we don't have to choose this time.

For now, our options are:
1. We leave everything as-is. The obvious drawback is that we leave this bug open. Actually, a whole family of bugs.
2. We make some functions non-`const`. This is one kind on inconvenience as they will not be usable within const expressions anymore, 
3. We make API fallible. This is another kind of inconvenience, as users will have to unwrap value one way or another. And there's no way to unwrap in const expressions.
4. We can wait for stabilization of `const_panic`. It is already in beta so, hopefully, it will soon be in stable. With `const_panic` we have two more options:
   1. Add functions for checked conversions and make existing API depend on them.
   2. Replace existing conversions with checked ones, as unwrapping will be possible within const expressions.

   The only drawback I see is that we'll have to bump MSRV. 

Personally, I'd pick 4-th option. Bumping MSRV for the sake of fixing bugs seems reasonable. 

[...]

PR #385, opened 2021-11-04
Fixed doctests for `OffsetDateTime::checked_{add,sub}`

L117:
Due to a bug in `PrimitiveDateTime::assume_offset` some doctests were passing, though they should not.
These test were replaced with correct ones.


jhpratt:
I think you misunderstood me in #384. My intent is to add an assertion in `OffsetDateTime::replace_offset`, as I'd previously decided that an underlying `PrimitiveDateTime` that is invalid if standalone is acceptable. These two doctests _should_ pass. It's just a matter of implementing the necessary checks to ensure standalone invalid values aren't exposed (a la `OffsetDateTime::replace_offset`).

PR #386, opened 2021-11-05
Fixed `OffsetDateTime::checked_{add,sub}` implementations

L117:
This fixes invalid behavior of aforementioned functions when internal `utc_datetime` has (intentional) out-or-range value.

### DIFF
--- a/src/offset_date_time.rs
+++ b/src/offset_date_time.rs
@@ -704,8 +704,8 @@ impl OffsetDateTime {
     /// );
     /// ```
     pub const fn checked_add(self, duration: Duration) -> Option<Self> {
-        let utc_datetime = const_try_opt!(self.utc_datetime.checked_add(duration));
-        Some(utc_datetime.assume_utc().to_offset(self.offset))
+        let offset_datetime = self.utc_datetime.utc_to_offset(self.offset);
+        Some(const_try_opt!(offset_datetime.checked_add(duration)).assume_offset(self.offset))
     }
 
     /// Computes `self - duration`, returning `None` if an overflow occurred.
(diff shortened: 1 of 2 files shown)


src/offset_date_time.rs
```diff
@@ -287,6 +287,29 @@ impl OffsetDateTime {
...
+        let time = Time::__from_hms_nanos_unchecked(
+            hour as _,
+            minute as _,
+            second as _,
+            self.nanosecond(),
+        );
+
+        PrimitiveDateTime::new(date, time)
+    }
+
```
jhpratt, line 312: There's `PrimitiveDateTime::utc_to_offset` and `PrimitiveDateTime::offset_to_utc`, which are internal methods that should be self-explanatory.

src/offset_date_time.rs
```diff
@@ -704,8 +727,10 @@ impl OffsetDateTime {
     /// );
     /// ```
     pub const fn checked_add(self, duration: Duration) -> Option<Self> {
-        let utc_datetime = const_try_opt!(self.utc_datetime.checked_add(duration));
-        Some(utc_datetime.assume_utc().to_offset(self.offset))
+        match self.date_time().checked_add(duration) {
+            Some(datetime) => Some(datetime.assume_offset(self.offset)),
+            None => None,
+        }
```
jhpratt, line 733:
```suggestion
        const_try_opt!(self.date_time().checked_add(duration)).assume_offset(self.offset)
```

jhpratt (changes requested):
LGTM with minor cleanups.

jhpratt (approved):
LGTM.

PR #391, opened 2021-11-09
Add example to rfc3339's docs

Shadlock0133:


jhpratt:
If we're going to add an example, why not have it be actual code showing how it's used?

Shadlock0133:
`Rfc3339` is a struct representing a format description, so I would expect an example of the format, and maybe a formatting string?
I think code example makes sense on [`parse`](https://docs.rs/time/0.3.4/time/struct.OffsetDateTime.html#method.parse)/[`format`](https://docs.rs/time/0.3.4/time/struct.OffsetDateTime.html#method.format) functions, which already have them. But maybe `Rfc3339` could have links, like:

```rust
/// The format described in [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6).
///
/// Format description: "[year]-[month]-[day]T[hour]:[minute]:[second].[subsecond]Z"
///
/// Example: 1985-04-12T23:20:50.52Z
///
/// Can be used with `parse` and `format` functions of [`Time`], [`Date`], ...
pub struct Rfc3339;
```

jhpratt:
Perhaps having some sort of indication that it can be passed to those methods is what I was envisioning. One thing we don't want, though, is the `"[year]-[month]-[day]T[hour]:[minute]:[second].[subsecond]Z"` comment — it's not accurate as there are nuances that make it different.

jhpratt:
LGTM. Thanks.

PR #401, opened 2021-11-25
feat: additional formatting options for human readable parsing

kate-shine:
Serialized times are often written in different human readable formats than strictly hh:mm:ss.ms

This also makes deserializer behavior closer to the time!() macro

### DIFF
--- a/src/serde/mod.rs
+++ b/src/serde/mod.rs
@@ -229,10 +229,14 @@ const TIME_FORMAT: &[FormatItem<'_>] = &[
     FormatItem::Component(Component::Hour(<modifier::Hour>::default())),
     FormatItem::Literal(b":"),
     FormatItem::Component(Component::Minute(<modifier::Minute>::default())),
-    FormatItem::Literal(b":"),
-    FormatItem::Component(Component::Second(<modifier::Second>::default())),
-    FormatItem::Literal(b"."),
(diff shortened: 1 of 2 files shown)


jhpratt:
This isn't something I'm interested in. The default deserializer is capable of parsing the format generated by the serializer, which is all that's necessary. Custom serializers/deserializers can be used with serde if you need something different. This will be made _much_ simpler with #381.

kate-shine:
> This isn't something I'm interested in. The default deserializer is capable of parsing the format generated by the serializer, which is all that's necessary. Custom serializers/deserializers can be used with serde if you need something different. This will be made _much_ simpler with #381.

I would understand this reasoning with the `serde` feature, but `serde-human-readable` quite implies that at lest the h:m:s format will be possible to read. It's also a common time format used in json. From my experience, inclusion of subseconds is really uncommon. While #381 indeed solves it, it will force people to write their own custom deserializers for common use cases.

jhpratt:
The format is the only one that includes all information. "Human readable" in this context means _not binary_ and _intended for human consumption_. The default deserializer handles the default serializer, which is all it needs to do. The intended use case for using the default is for communication between various things using this crate. If you need to communicate with something else, you should be explicit in the format that is handled.

Writing a custom deserializer will be far easier than you think when the aforementioned PR lands.

PR #403, opened 2021-11-25
Add RFC2822 well-known format

### LINKED ISSUE #399: Add RFC2822 well-known format description
RFC2822 formatted dates are used in many places in web contexts, one of the more noteworthy of those being `Date` headers in the HTTP protocol. An example of a date in this format could be `Fri, 21 Nov 1997 09:55:06 -0600`.

As an additional note, the `chrono` crate has 2822 as a well known format, and adding it to `time` could be a big help to those interesting in migrating to `time`.

[The RFC2822 specification](https://datatracker.ietf.org/doc/html/rfc2822)

agraven:
Closes #399.

I've tried to stick to the robustness principle to a reasonable degree in the implementation, and included handling for obsolete parts of the RFC which seemed likely to be encountered.

<del>One of the doc tests fails with an error about the `Rfc2822` struct not implementing `Deref` which I'm not able to make sense of.</del> fixed

### DIFF
--- a/src/formatting/formattable.rs
+++ b/src/formatting/formattable.rs
@@ -3,9 +3,11 @@
 use core::ops::Deref;
 use std::io;
 
-use crate::format_description::well_known::Rfc3339;
+use crate::format_description::well_known::{Rfc2822, Rfc3339};
 use crate::format_description::FormatItem;
-use crate::formatting::{format_component, format_number_pad_zero, write};
+use crate::formatting::{
+    format_component, format_number_pad_zero, write, MONTH_NAMES, WEEKDAY_NAMES,
+};
 use crate::{error, Date, Time, UtcOffset};
 
 /// A type that can be formatted.
@@ -14,6 +16,7 @@ pub trait Formattable: sealed::Sealed {}
 impl Formattable for FormatItem<'_> {}
 impl Formattable for [FormatItem<'_>] {}
 impl Formattable for Rfc3339 {}
+impl Formattable for Rfc2822 {}
 impl<T: Deref> Formattable for T where T::Target: Formattable {}
 
 /// Seal the trait to prevent downstream users from implementing it.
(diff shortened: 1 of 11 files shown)


agraven:
I figured out the issue, I had forgotten to add `impl Parsable for Rfc2822 {}`. I also fixed some oversights in the parsing and added some tests.

agraven:
I added tests for the formatting as well.

src/formatting/formattable.rs
```diff
@@ -102,6 +105,60 @@ where
...
+
+        let mut bytes = 0;
+
+        let year = date.year();
+
+        if year < 1900 {
+            return Err(error::Format::InvalidComponent("year"));
+        }
+        if time.nanosecond() != 0 {
+            return Err(error::Format::InvalidComponent("nanosecond"));
```
agraven, line 128: Not sure whether to fail on non-zero nanoseconds or not, RFC2822 can't represent non-zero nanoseconds, but failing formatting on those seems like big potential for a footgun that would take library consumers by surprise. E.g. `OffsetDateTime::now_utc().format(&Rfc2822)?` would error in 99.999% of cases.

jhpratt: I think it is reasonable to ignore if nonzero. That's what happens with manually created formats unless it's explicitly provided, so it would be consistent.

agraven: Will fixup e257222 into 92d1a69 in that case

jhpratt (changes requested):
I haven't yet looked over the specification for compliance purposes, but the implementation looks to be in solid shape. Just a couple minor nits on the implementation that I've left inline.

src/parsing/parsable.rs
```diff
@@ -112,6 +113,124 @@ where
...
+        if let Some(zone_literal) = zone_literal {
+            let input = zone_literal.assign_value_to(&mut parsed.offset_hour);
+            parsed.offset_minute = Some(0);
+            return Ok(input);
+        }
+
+        let ParsedItem(input, offset_sign) = sign(input).ok_or(InvalidComponent("offset hour"))?;
+        let input = exactly_n_digits::<u8, 2>(input)
+            .ok_or(InvalidComponent("offset hour"))?
+            .map(|offset_hour| match offset_sign == b'-' {
```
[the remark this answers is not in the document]
agraven: Done

agraven:
I addressed the review feedback.

jhpratt:
Starting to check for spec compliance. Apparently RFC2822 was obsoleted by RFC5322. I'll see if the datetime formats are the same and if so will re-export the struct.

If there's any minute details from the spec that haven't been implemented, I think I can handle it unless you'd prefer to for whatever reason. Thank you for your contribution!

agraven:
You doing the rest is fine by me.

jhpratt:
Alright. Managed to parse this according to the spec. Needed to introduce a bunch of new combinators to do so because parsing "whitespace" involves a whole lot more than just doing that. Once I get the test coverage back up I'll merge.

jhpratt:
I believe this PR is in sufficient condition to merge now, and just need to make a final pass to make sure I didn't miss anything. The `cfws` rule is stupidly complex for no apparent reason. Diff coverage is 100%.

jhpratt (approved):

jhpratt:
Thank you once again!

Issue #375, opened 2021-10-26
`format_into()` requires `io::Write`; what about `fmt::Write` in Debug/Display impls?


aldanor:
Simple (and the most common, I believe) use case - you're implementing `Display` or `Debug` for your type containing some time objects, and you are provided with [`std::fmt::Formatter`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html) which [implements](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#impl-Write) [`std::fmt::Write`](https://doc.rust-lang.org/std/fmt/trait.Write.html) trait. (maybe I'm missing something trivial, so apologies in advance if that's the case)

Current `format_into` implementations though [require](https://docs.rs/time/0.3.4/time/struct.OffsetDateTime.html#method.format_into) [`&mut impl std::io::Write`](https://doc.rust-lang.org/std/io/trait.Write.html) which is a different beast, more complex and aimed at different stuff, with vectored writes, flushing and all the rest. If one were to choose between the two, wouldn't it make sense to support `fmt::Write` and not `io::Write`? Is there a clean alternative way?

(There's crates like [`fmt2io`](https://docs.rs/fmt2io/0.2.0/fmt2io) that provide hacky bridges but that's just ugly and wrong)

mzabaluev:
It would be nice to make `format` (which only requires `alloc`) and any future formatting methods using `core::fmt::Write` available in no-std. However, the implementation and feature gate for `format` is coupled with `format_into` which depends on `std::io::Write`.

To summarize, the three use cases can all be made available behind different feature gates:

| Method | Features |
| ---------- | ----------------- |
| `format` | `alloc` |
| `format_into` | `std` |
| `format_into_fmt_writer` | |

jhpratt:
The primary problem that formatting into `fmt::Write` is that arbitrary bytes can be present in the format description. It's not immediately clear how this interacts with an output that requires UTF-8.

That said, the existing `format` method _can not_ change as it would be breaking to do so.

mzabaluev:
> That said, the existing `format` method _can not_ change as it would be breaking to do so.

No, but it could be made available under a `formatting-alloc` feature gate that would only imply `alloc`, and would be auto-enabled by full `formatting`.

jhpratt:
It has to be under a different method name. Feature gates can't change a method signature.

mzabaluev:
> Feature gates can't change a method signature.

[...]

PR #407, opened 2021-12-06
Add backwards compatible support for `fmt::Write`

Mathspy:
Hello, first thank you for all the work on this lovely library! ❤️ 

I am submitting this PR for discussion and learning more than I am trying to get it actually merged

# Implementation
This resolves https://github.com/time-rs/time/issues/375 by adding a compatibility layer between `fmt::Write` and `io::Write`. And switching the formatting machinery to use `fmt::Write`.

# Rationale
The reason for picking `fmt::Write` as the canonical implementation is because of [this quote from Rust `fmt::Write` doc page](https://doc.rust-lang.org/std/fmt/trait.Write.html):
> This trait only accepts UTF-8–encoded data and is not flushable. If you only want to accept Unicode and you don’t need flushing, you should implement this trait; otherwise you should implement `std::io::Write`.

Since we only want to accept UTF-8 (Is this assumption correct?) and we don't need flushing (couldn't find any calls to `.flush()` in the codebase) it felt more correct to use `fmt::Write`.
Also since `fmt::Write` has more guarantees (i.e UTF-8) it's easier to implement `fmt::Write` for `io::Write` than the other way around.

# Uncertainties 
- I wasn't really sure what to call the new methods so I went with this suggestion https://github.com/time-rs/time/issues/375#issuecomment-978872293.
- Same thing for the sealed internal methods, I went with renaming to `format_into_old` but I can keep the names of internal methods matching the names of externals methods.

# Possible follow-ups
- We can add a `SafeLiteral(&str)` to `FormatItem`. We should also be able to update the macros and runtime parsers to use that instead of `Literal(&[u8])`.
- If you think this is PR is a good idea, it's up to you to decide whether you want to keep support for both `Write`s for foreseeable future or to deprecate the support for `io::Write` to be removed in a future breaking change. (Could also deprecate `Literal` in favor of `SafeLiteral` in parallel)
--------
Tell me what do you think and if you think this is taking a completely wrong dimension please don't hesitate to close it!

Thank you ❤️

### DIFF
--- a/src/formatting/formattable.rs
+++ b/src/formatting/formattable.rs
@@ -135,6 +135,26 @@ impl<'a> sealed::Sealed for FormatItem<'a> {
 
         Ok(())
     }
+
+    fn format_into_old(
+        &self,
+        output: &mut impl io::Write,
+        date: Option<Date>,
+        time: Option<Time>,
+        offset: Option<UtcOffset>,
+    ) -> Result<usize, error::Format> {
+        match self {
+            Self::Literal(literal) => {
+                output.write_all(literal)?;
+                Ok(literal.len())
+            }
+            other => {
+                let mut compat = Compat::from_io(output);
+                let result = other.format_into(&mut compat, date, time, offset);
+                compat.into_io_result(result)
+            }
+        }
+    }
 }
 
 impl<'a> sealed::Sealed for [FormatItem<'a>] {
--- a/src/formatting/formattable.rs
+++ b/src/formatting/formattable.rs
@@ -48,14 +48,37 @@ mod sealed {
     /// Format the item using a format description, the intended output, and the various components.
     #[cfg_attr(__time_03_docs, doc(cfg(feature = "formatting")))]
     pub trait Sealed {
-        /// Format the item into the provided output, returning the number of bytes written.
+        /// Format the item into the provided output.
         fn format_into(
+            &self,
+            output: &mut impl fmt::Write,
+            date: Option<Date>,
+            time: Option<Time>,
+            offset: Option<UtcOffset>,
+        ) -> Result<(), error::Format>;
+
+        /// Format the item into the provided output, returning the number of bytes written.
+        fn format_into_old(
             &self,
             output: &mut impl io::Write,
             date: Option<Date>,
             time: Option<Time>,
             offset: Option<UtcOffset>,
-        ) -> Result<usize, error::Format>;
+        ) -> Result<usize, error::Format> {
+            let mut compat = Compat {
+                writer: output,
+                bytes_written: 0,
+                error: None,
+            };
+            self.format_into(&mut compat, date, time, offset)
+                .map(|_| compat.bytes_written)
+                .map_err(|fmt_error| {
+                    compat
+                        .error
+                        .map(error::Format::from)
+                        .unwrap_or_else(|| error::Format::from(fmt_error))
+                })
+        }
 
         /// Format the item directly to a `String`.
         fn format(
@@ -64,9 +87,9 @@ mod sealed {
             time: Option<Time>,
             offset: Option<UtcOffset>,
         ) -> Result<String, error::Format> {
-            let mut buf = Vec::new();
+            let mut buf = String::new();
             self.format_into(&mut buf, date, time, offset)?;
-            Ok(String::from_utf8_lossy(&buf).into_owned())
+            Ok(buf)
         }
     }
 }
--- a/src/formatting/formattable.rs
+++ b/src/formatting/formattable.rs
@@ -19,9 +19,13 @@ impl Formattable for Rfc3339 {}
 impl Formattable for Rfc2822 {}
 impl<T: Deref> Formattable for T where T::Target: Formattable {}
 
+/// A compatibility layer to translate [`io::Write`] into [`fmt::Write`]
 struct Compat<'a, W: io::Write> {
+    /// The [`io::Write`]r to apply the translation on
     writer: &'a mut W,
+    /// The total bytes written into the writer so far
     bytes_written: usize,
+    /// The last error from the writer if it returned errors
     error: Option<io::Error>,
 }
 
@@ -76,7 +80,7 @@ mod sealed {
                     compat
                         .error
                         .map(error::Format::from)
-                        .unwrap_or_else(|| error::Format::from(fmt_error))
+                        .unwrap_or_else(|| fmt_error)
                 })
         }
 
@@ -103,7 +107,7 @@ impl<'a> sealed::Sealed for FormatItem<'a> {
         time: Option<Time>,
         offset: Option<UtcOffset>,
     ) -> Result<(), error::Format> {
-        Ok(match *self {
+        match *self {
             Self::Literal(literal) => {
                 output.write_str(String::from_utf8_lossy(literal).as_ref())?
             }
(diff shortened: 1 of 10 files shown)


jhpratt:
Hey there. Sorry about the delay in getting to this. Would you mind rebasing? The conflict is just due to a dependency upgrade and is trivial. The other thing I immediately see is that the error branches aren't covered; could you do this? `std::fmt::Error` is publicly a ZST so it shouldn't be difficult to construct a case that covers this.

Once these two things are done I'll do an in-depth review and merge if appropriate.

jhpratt:
Oh, and assuming that the bytes are UTF-8 is _not_ an assumption that should be made. It is currently possible to construct a `FormatDescription` manually with raw bytes, and I believe that there are tests that rely on this. I'm 99% certain there are real-world users that need this behavior too.

Mathspy:
> Hey there. Sorry about the delay in getting to this. Would you mind rebasing? The conflict is just due to a dependency upgrade and is trivial. The other thing I immediately see is that the error branches aren't covered; could you do this? `std::fmt::Error` is publicly a ZST so it shouldn't be difficult to construct a case that covers this.
> 
> Once these two things are done I'll do an in-depth review and merge if appropriate.

No worries! Happy new year ❤️ 
Fixed conflict and covered fixed coverage

> Oh, and assuming that the bytes are UTF-8 is _not_ an assumption that should be made. It is currently possible to construct a `FormatDescription` manually with raw bytes, and I believe that there are tests that rely on this. I'm 99% certain there are real-world users that need this behavior too.

Yeah it looks like my assumption was indeed in correct, I was able to observe different behaviors using this test case

```rs
use time::format_description::FormatItem;
use time::macros::date;

const NON_UTF8: &[FormatItem<'_>] = &[
    FormatItem::Literal(&[195, 40]),
];

fn main() {
    let today = date!(2022 - 01 - 03);

    let mut bytes = vec![195, 40];

    dbg!(today.format_into(&mut bytes, NON_UTF8));
    dbg!(bytes);
}
```

On `main` branch this produced `[195, 40, 195, 40]` but on this branch it produces `[195, 40, 239, 191, 189, 40]` so this isn't mergable as-is.

One possible solution is to reverse the roles and implement `io::Write` for `fmt::Write` like this:

```rs
struct Compat<'a, W: fmt::Write> {
    writer: &'a mut W,
}

impl<'a, W> io::Write for Compat<'a, W>
where
    W: fmt::Write,
{
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let string = String::from_utf8_lossy(buf);

        self.writer
            .write_str(&string)
            .map(|_| string.len())
            .map_err(|_| io::Error::new(io::ErrorKind::Other, "Internal fmt error"))
    }

    fn flush(&mut self) -> io::Result<()> {
        Ok(())
    }
}
```

But this can be done in userland so I am not really sure if there's any benefit to including it here. 
One other possible solution is for the `Sealed` trait to require a manual implementation for `format_into` and `format_into_fmt_write` where they'd be pretty much identical implementations except for behavior around `FormatItem::Literal` but that seems like a lot of duplicate code to keep around.

So I am not sure if there's anything here that's actionable for `time` crate anymore, should I close this?

src/formatting/formattable.rs
```diff
@@ -54,37 +102,40 @@ mod sealed {
...
         date: Option<Date>,
         time: Option<Time>,
         offset: Option<UtcOffset>,
-    ) -> Result<usize, error::Format> {
-        Ok(match *self {
-            Self::Literal(literal) => write(output, literal)?,
+    ) -> Result<(), error::Format> {
+        match *self {
+            Self::Literal(literal) => {
+                output.write_str(String::from_utf8_lossy(literal).as_ref())?
```
Mathspy, line 112: This `from_utf8_lossy` call is where my assumption about utf-8 falls through

jhpratt:
> But this can be done in userland so I am not really sure if there's any benefit to including it here.

Well, it's been requested a few times so there's _some_ benefit in just doing it once so others don't have to. It's not like it's out of scope or anything.

> One other possible solution is for the Sealed trait to require a manual implementation for format_into and format_into_fmt_write where they'd be pretty much identical implementations except for behavior around FormatItem::Literal but that seems like a lot of duplicate code to keep around.

Wouldn't the only place that would need to be changed be `impl<'a> sealed::Sealed for FormatItem<'a>`? That is the implementation that everything ultimately falls back on. If this is the case it's not that much code to duplicate: just a dozen lines or so. Well-known types can be passed through `String::from_utf8` (matching and using `unreachable!("…")` on the result), as the formats are known to be UTF-8 in their entirety.

My thought right now (having largely not reviewed the approach you actually took) is to create a private trait that any type implementing `io::Write` or `fmt::Write` implements. All internal APIs can then be made generic over that trait. Right off the top of my head I think this might run into a currently unsolvable problem, though: a type can implement both `io::Write` and `fmt::Write` and specialization is not stable. If this is in fact the case, then altering the aforementioned implementation is probably the way to go unless I'm missing something?

Mathspy:
> Wouldn't the only place that would need to be changed be `impl<'a> sealed::Sealed for FormatItem<'a>`? That is the implementation that everything ultimately falls back on. If this is the case it's not that much code to duplicate: just a dozen lines or so. Well-known types can be passed through `String::from_utf8` (matching and using `unreachable!("…")` on the result), as the formats are known to be UTF-8 in their entirety.

This made me realize that there was actually a path forward within the current implementation, I pushed it in the latest commits alongside a test to verify that it works (and that it indeed did fail before the changes)

> My thought right now (having largely not reviewed the approach you actually took) is to create a private trait that any type implementing `io::Write` or `fmt::Write` implements. All internal APIs can then be made generic over that trait. Right off the top of my head I think this might run into a currently unsolvable problem, though: a type can implement both `io::Write` and `fmt::Write` and specialization is not stable.

I am not entirely sure how would this look, could you elaborate on it? I agree that this sounds like something that would quickly hit specialization though

src/formatting/formattable.rs
Mathspy: Fixed in https://github.com/time-rs/time/pull/407/commits/959bb7399d3befee840af6f0338a886e87615468

jhpratt:
> I am not entirely sure how would this look, could you elaborate on it? 

Basically it would require all internal methods like `format_number` to be generic over a trait that shims the two writers into a desired API. I'm quite certain it's not currently possibly, though.

I'll take a look at the code soon™

jhpratt:
Finally coming back around to this. After some more thought, I don't think that this is the best way forward. Duplicating methods has some overhead from the user's perspective, as you need to know which to call — and they're split for no apparent reason to someone new to or unfamiliar with Rust. In addition, the compatibility is ultimately a hack. While possible to avoid, doing so would lead to significant duplication as noted earlier in this thread.

For these reasons (along with minor nits), I'm going to close this. I sincerely thank you for your work on this; it's not trivial to do. Your work is appreciated even though this PR is not ultimately being merged.

Looking forward, I think specialization will lead to the ideal solution here. Being able to accept either `io::Write` or `fmt::Write` is clearly preferable, and having the ability to prefer `io::Write` when a parameter implements both would be great. This would be wholly transparent to the user, and would avoid some of the concerns introduced here. To my knowledge `min_specialization` is sound and may be stabilized later this year. Full specialization is not needed here.

Mathspy:
Sounds good! And thank you for your time ❤️ Looking forward to specialization stabilization as well

PR #412, opened 2021-12-09
Tighten parsing of leap second times in RFC 3339 and RFC 2822

### LINKED ISSUE #406: Tighten leap second handling when parsing from RFC 3339
The current implementation of `format_description::well_known::Rfc3339` allows second 60 to occur in any time, which is then rolled back to the last nanosecond of 59. As leap seconds are only really allowed at 23:59:60 UTC, this behaviour feels way too loose.

Fortunately, there is always the full parsed date for the `Rfc3339` format parser to look at. So it could compute the UTC time and check if the hour is 23 and the minute is 59, otherwise it should reject the parsed time. It might make sense to also check if it's the last day of the month, though I don't know if the language in ITU-R TF.460-6 is considered normative by the IERS.

mzabaluev:
Resolves #406

### DIFF
--- a/src/error/try_from_parsed.rs
+++ b/src/error/try_from_parsed.rs
@@ -15,6 +15,8 @@ pub enum TryFromParsed {
     InsufficientInformation,
     /// Some component contained an invalid value for the type.
     ComponentRange(error::ComponentRange),
+    /// Leap second is not allowed to occur at this date and time.
+    LeapSecondNotValid,
 }
 
 impl fmt::Display for TryFromParsed {
@@ -24,6 +26,9 @@ impl fmt::Display for TryFromParsed {
                 "the `Parsed` struct did not include enough information to construct the type",
             ),
             Self::ComponentRange(err) => err.fmt(f),
+            Self::LeapSecondNotValid => {
+                f.write_str("leap second is not allowed to occur at this date and time")
+            }
         }
     }
 }
@@ -51,6 +56,7 @@ impl std::error::Error for TryFromParsed {
         match self {
             Self::InsufficientInformation => None,
             Self::ComponentRange(err) => Some(err),
+            Self::LeapSecondNotValid => None,
         }
     }
 }
--- a/src/error/try_from_parsed.rs
+++ b/src/error/try_from_parsed.rs
@@ -15,8 +15,6 @@ pub enum TryFromParsed {
     InsufficientInformation,
     /// Some component contained an invalid value for the type.
     ComponentRange(error::ComponentRange),
-    /// Leap second is not allowed to occur at this date and time.
-    LeapSecondNotValid,
 }
 
 impl fmt::Display for TryFromParsed {
@@ -26,9 +24,6 @@ impl fmt::Display for TryFromParsed {
                 "the `Parsed` struct did not include enough information to construct the type",
             ),
             Self::ComponentRange(err) => err.fmt(f),
-            Self::LeapSecondNotValid => {
-                f.write_str("leap second is not allowed to occur at this date and time")
-            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/error/try_from_parsed.rs
```diff
@@ -15,6 +15,8 @@ pub enum TryFromParsed {
     InsufficientInformation,
     /// Some component contained an invalid value for the type.
     ComponentRange(error::ComponentRange),
+    /// Leap second is not allowed to occur at this date and time.
+    LeapSecondNotValid,
```
jhpratt, line 19: Why not reuse the `ComponentRange` error type & variant? It can handle this quite easily.

src/parsing/parsable.rs
```diff
@@ -172,8 +172,10 @@ impl sealed::Sealed for Rfc3339 {
         };
 
         // The RFC explicitly allows leap seconds. We don't currently support them, so treat it as
-        // the previous moment.
+        // the nearest preceding moment that can be represented, but set a flag to perform
+        // special validation when the time value is constructed.
         if parsed.second == Some(60) {
+            parsed.leap_second_input = true;
```
jhpratt, line 178: In my opinion a better approach would be simply set the second to 60 unconditionally and only check whether it's a leap second in the `TryFrom` impl. Then the only thing that would be necessary is knowing whether the format is well-known (which can be set unconditionally).

I am aware that this is currently permitted on the latest release, but this could certainly be considered a bug fix due to the lack of information in the final value. It's a mistake on my end, admittedly.

src/parsing/parsed.rs
```diff
@@ -648,31 +655,44 @@ impl TryFrom<Parsed> for Date {
     }
 }
 
+fn naive_time_from_parsed(parsed: Parsed) -> Result<Time, error::TryFromParsed> {
```
jhpratt, line 658: While I understand what you're going for here (and other similar locations), it would probably be best to reject the leap second outright for types other than `OffsetDateTime`, as they don't contain the necessary information in the final value to know if it was at a leap second or not.

jhpratt (changes requested):
Overall looks good. The primary thing here is delaying the check to the `TryFrom` impl, which would have the additional benefit of not requiring special handling for each well-known format (the format just needs to set a flag that it accepts leap seconds). Other than that, removing the code that I inserted without thinking would be great (this is noted inline as well).

src/error/try_from_parsed.rs
mzabaluev: It feels a bit cryptic to fail second 60 as a plain range error when its validity actually depends on the content of other components.
The user might be interested to know why their input has failed if it was meant to be a leap second indeed. With an offset, it may not be obvious that the time is not actually at 23:59 UTC.

src/parsing/parsable.rs
mzabaluev: OK, will redo as proposed.

src/parsing/parsed.rs
mzabaluev: The information is available in the RFC 3339 source data to make that call. But if it isn't, e.g. when the `Parsed` is coming from another well-known format, then it's an "insufficient information" error. I agree this could be cryptic. Will redo as proposed.

src/error/try_from_parsed.rs
jhpratt: There is a "conditional" field (or something with a similar name; I'm typing this off-hand) in the struct for things like this. A similar issue arises when validating the number of days in a month/year.

jhpratt:
Hey @mzabaluev, sorry for the delay. I just reviewed this PR and added a commit with minor changes locally. I'm unable to push it though. Have you checked the box that permits edits from the maintainers (a.k.a. me)?

jhpratt:
Disregard. I was trying to push to a branch of a different name.

jhpratt:
Clippy failure is unrelated. Merging. Thank you!

PR #419, opened 2021-12-30
Allow macOS to get local timezone offset if single threaded

ephraimkunz:
Recently support for Linux and BSD was merged to allow code running on those OS's to access the local timezone if code is single threaded. This simply adds support for macOS. `num_threads` already supports macOS so this just enables it in `time`.


jhpratt:
This is already done on the `dev` branch.

PR #451, opened 2022-03-08
Use tz-rs to obtain the system's UTC offset

clia:
If this is an annoying PR, please ignore it!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ time-macros = { version = "=0.2.3", path = "time-macros", optional = true }
 [target.'cfg(target_family = "unix")'.dependencies]
 libc = "0.2.98"
 num_threads = "0.1.2"
+tz-rs = "0.6.5"
 
 [dev-dependencies]
 rand = { version = "0.8.4", default-features = false }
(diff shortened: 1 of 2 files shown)


jhpratt:
This is something I planned to look into at some point. It's a _very_ large change and will necessitate me doing an in-depth review of the full `tz-rs` crate. Setting that aside, it looks like the crate requires a newer version of Rust that `time` cannot yet use per policy. Closing for that reason.

clia:
It's OK! We all support your hard work!

For now I made a temporary crate [clia-time](https://crates.io/crates/clia-time) for myself and will have a try.

jhpratt:
Please update the README to indicate that there is no relation to this crate.

clia:
Updated.

x-hgg-x:
For information, I have reduced the MSRV to `rustc 1.45` for the version `v0.6.6` of `tz-rs`.

jhpratt:
Awesome. I'll take a look at some point. Ultimately it's still probably unlikely because it's such a large change, plus the tz-rs crate doesn't seem too stable. Rust is getting close to a satisfactory solution that would permit me to soundly expose the existing functionality.

PR #454, opened 2022-03-14
OffsetDateTime: Add conversions to PrimitiveDateTime

uklotzde:
A `PrimitiveDateTime` can be converted conveniently into `OffsetDateTime` with `assume_utc()` and `assume_offset()`. What is currently missing are conversion functions in the opposite direction.

This PR adds the missing functions to `OffsetDateTime` to accomplish these conversions efficiently. It is doable with the public 0.3.7 API but would require intermediate conversions which might not be optimized away.

The naming is descriptive yet verbose and could be improved.

### DIFF
--- a/src/offset_date_time.rs
+++ b/src/offset_date_time.rs
@@ -107,6 +107,41 @@ impl OffsetDateTime {
         }
     }
 
+    /// Obtain the corresponding UTC timestamp without offset as [`PrimitiveDateTime`].
+    ///
+    /// ```rust
+    /// # use time::macros::datetime;
+    /// assert_eq!(
+    ///     datetime!(2018-12-31 23:00:00),
+    ///     datetime!(2019-01-01 01:00:00 +02:00).to_primitive_utc(),
+    /// );
+    /// ```
+    pub fn to_primitive_utc(self) -> PrimitiveDateTime {
+        let Self {
+            utc_datetime,
+            offset: _,
+        } = self;
+        utc_datetime
+    }
+
+    /// Obtain the corresponding timestamp including the UTC offset as [`PrimitiveDateTime`].
+    ///
+    /// ```rust
+    /// # use time::macros::datetime;
+    /// assert_eq!(
+    ///     datetime!(2019-01-01 01:00:00),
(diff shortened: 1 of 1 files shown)


jhpratt:
This is an intentional omission. Including something like this would make it far too easy for people to screw something up without realizing it. Explicit handling of offsets is nearly always best.

uklotzde:
Unfortunately, this conversion is still needed for some use cases with an implicit offset. It requires unneeded code just because the wrapped `PrimitiveDateTime` is inaccessible.

jhpratt:
As I said, explicitly handling of offsets is nearly always best. The "implicit offset" you mention is difficult to get correct.

uklotzde:
Considering your arguments then what's the point of providing `PrimitiveDateTime` at all with an incomplete, asymmetric set of operations that don't support a round-trip? Though I agree that all conversions need to be explicit, especially those that are *lossy* and drop information.

`PrimitiveDateTime` should mostly be used as a *view* as [chrono](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.naive_utc) also states for obtaining the corresponding `NaiveDateTime`. A complete and consistent set of conversion functions still makes sense to me.

jhpratt:
`PrimitiveDateTime` exists because an offset is not always known. It doesn't mean that it's a fixed offset that just so happens to be omitted.

While `time` took some inspiration from `chrono` in the past, it's by no means a guiding principle.

uklotzde:
After my co-worker @flosse stumbled over the need for such a conversion in another project I remembered when I once needed it.

You asked for use cases? Here is just one of them in a popular Rust project:

https://github.com/quickwit-oss/tantivy/blob/23fe73a6c0a0a9f9b0a263f7fa31b278a9a6101f/src/lib.rs#L225

Whenever you want to map time stamps to a lower level transport or storage layer you need to strip off the time zone information somehow and manage it separately. Either explicitly or, as in this case, implicitly or by convention. This requires a bidirectional mapping without revealing any internal details of this library. Don't only think in application-level uses cases, there are also technical use cases.

As a consequence of denying these use cases everyone is forced to implement their custom conversions again and again.

Issue #395, opened 2021-11-14
Getting the `PrimitiveDateTime` (`utc_datetime`) component of `OffsetDateTime`?


Twixes:
Hey! In my application I want to expose two timestamp data types: one with time zone information and one without (assumed to be UTC). I thought that the `OffsetDateTime` and `PrimitiveDateTime` structs respectively would fit this model ideally, but I haven't found a way to access `OffsetDateTime`'s `utc_datetime` field (which is the `PrimitiveDateTime` backing it), and there's no `PrimitiveDateTime::now_utc()` either, so it seems that there's actually no way for me to use `PrimitiveDateTime`.
I can wrap `OffsetDateTime` in a `PrimitiveDateTime`-like newtype for differentiation (e.g. so that my serialization/deserialization trait implementation knows when it's needed to encode `offset`)… but that'd be a lot of boilerplate.
Why not just expose `utc_datetime`? Or am I missing something?

Twixes:
Actually scratch that. I see that `OffsetDateTime` includes many useful methods that `PrimitiveDateTime` doesn't, so the practicality of the latter is really limited anyway. I'll work around this using `OffsetDateTime` only.

jhpratt:
Yeah, the two structs are quite different (and for good reason). This has been requested a couple times recently, but I feel that using `.date()` and `.time()` are sufficient for those that want to do this. The internal representation isn't guaranteed, either.

Issue #390, opened 2021-11-09
Convert OffsetDateTime back to PrimitiveDateTime


DenisKolodin:
I know there are some discussions about converting `OffsetDateTime` to `PrimitiveDateTime`, but there is no a way to do that simply.

Since `chrono` used old version of `time` some projects considered to switch directly to the latest `time` crate, but there is a problem: `chrono` provides `naive_local` and `naive_utc` [methods](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#method.naive_utc), and it's not easy to use `time` directly if the project uses `NaiveDateTime` and already stored some data using that format (without timezone).

Is there any reason we don't have `From<OffsetDateTime> for PrimitiveDateTime` implementation? It could help to move from the `chrono` crate easier.

jhpratt:
> if the project uses `NaiveDateTime` and already stored some data using that format (without timezone).

That's the problem. You're trying to use `PrimitiveDateTime` for something where you should be using `OffsetDateTime`. `impl From<OffsetDateTime> for PrimitiveDateTime` won't be added in any situation because there is no `UtcOffset` that is always correct.

This was recently requested in #369. There is no reason that I see that you can't use `.date()` and `.time()`, recombining them into a `PrimitiveDateTime` if that's what you truly want. In #369 benchmarks were raised, but those were flawed. These methods aren't guaranteed to be free, regardless.

DenisKolodin:
@jhpratt Thanks 🙂

Issue #369, opened 2021-10-23
Split `OffsetDateTime` back into `PrimitiveDateTime` and an offset?


novacrazy:
Is there a way to split `OffsetDateTime` back into `PrimitiveDateTime` and an offset?

jhpratt:
I am dismissing the benchmarks as flawed, and I don't feel the desired use case and motivation aligns with the guarantees of the time crate (namely the internal representation of `OffsetDateTime`). As such I'm inclined to close this.

PR #458, opened 2022-04-11
Allow getting the PrimitiveDateTime of a OffsetDateTime

gunnarpieter:
This is already possible through getting the date and time and making a primitive but would just be nicer to have easy API.

### DIFF
--- a/src/offset_date_time.rs
+++ b/src/offset_date_time.rs
@@ -204,6 +204,19 @@ impl OffsetDateTime {
     pub const fn offset(self) -> UtcOffset {
         self.offset
     }
+    
+    
+    // region: getters
+    /// Get the [`PrimitiveDateTime`].
+    ///
+    /// ```rust
+    /// # use time::macros::{datetime};
+    /// assert_eq!(datetime!(2019-01-01 0:00 UTC).primitive_date_time(), datetime!(2019-01-01 0:00));
+    /// assert_eq!(datetime!(2019-01-01 0:00 +1).primitive_date_time(), datetime!(2019-01-01 0:00));
+    /// ```
+    pub const fn primitive_date_time(self) -> PrimitiveDateTime {
+        self.utc_datetime
+    }
 
     /// Get the [Unix timestamp](https://en.wikipedia.org/wiki/Unix_time).
     ///


jhpratt:
This was previously declined in #369, #390, #395, and #454. It is highly recommended to search before opening an issue or PR.

gunnarpieter:
> This was previously declined in #369, #390, #395, and #454. It is highly recommended to search before opening an issue or PR.

Why would this not be allowed? Obviously many people would like to do this. It is only logical as for many instances you do not want to have any Offset at all. Just the pure UTC and this will enforce only UTC by type.

Besides that there is also no way that new() is as fast as copying the underlying data structure directly.

jhpratt:
Please read the linked issues. I have repeatedly articulated my reasons for declining this request and will not do so again here.

gunnarpieter:
> Please read the linked issues. I have repeatedly articulated my reasons for declining this request and will not do so again here.

I have read those comments. Fine that you wish that people be safe with Time and 'most' people need the offset.
That does not mean all people need the offset. This is just a lacking API. 

The argument of internal representation changing is flawed due to the fact that PrimitiveDateTime is also a public struct and thus using date and time to create it will then likely also be broken due to this internal restructuring.

jhpratt:
People want the `utc_datetime` field exposed largely _because_ it's a field. That's terrible reasoning.

You want to discard information that is absolutely necessary. I have yet to see a use case where this API is actually needed. To give you an idea of how easy it is to get wrong, your comment is wrong.

> This is already possible through getting the date and time and making a primitive

This is not what you're doing in the PR. Time is complicated, and you should _never_ discard information you have about the offset. Doing so is extremely error prone.

Unless you _truly_ have something new to add, I'm not going to keep responding. I've been over this before.

gunnarpieter:
> This is not what you're doing in the PR. Time is complicated, and you should _never_ discard information you have about the offset. Doing so is extremely error prone.

Are you saying that creating an OffsetDateTime using `OffsetDateTime::now_utc()` then getting date and time and creating a PrimitiveDateTime is not the same as my PR?

Please enlighten me on why it is not the same, cause i would love to know...

> Doing so is extremely error prone.
Yeah only when you actually _need_ the offset. If you actually do not _need_ the offset it is not. If you read my comment above time passing is not different in UTC or +5 offsets.

jhpratt:
If you need the time passed, you should use `Duration::span` or `Instant`. Anything else is error-prone. Clocks aren't guaranteed to tick evenly, or even move forward in time.

If you call `now_utc`, it is the same. In **any** situation where the offset is not UTC the result would not be the same.

Stop calling my arguments "flawed". I have clearly articulated my reasons. I have far more experience with time handling and my goal is to reduce the amount of things _you_ do wrong. Not implementing this request is part of doing that.

gunnarpieter:
> If you call `now_utc`, it is the same. In **any** situation where the offset is not UTC the result would not be the same.

I am well aware. Thank you! I only want to do time arithmetic in the UTC timezone so keeping a timezone is useless.

> Stop calling my arguments "flawed".

Well, sadly they just are partially _flawed_. You can not use Duration if the Duration is yet unknown. I can not use Instant as it does not provide the full functionality required to do complex calculations and your offset eats into my limited embedded memory.

If i had plenty to spare I would really be fine with your view and technically I truly understand it. But this just wastes compute cycles and memory.

Anyways, I do not think we will figure this out. Which tbh, is kinda sad as your crate is all around the best related to time.
Just think about it and someday you might change your view. No hard feelings so, have a good evening/day!

jhpratt:
> I only want to do time arithmetic in the UTC timezone so keeping a timezone is useless.

It's semantic information that is certainly not useless.

> You can not use Duration if the Duration is yet unknown

I mentioned `Duration::time_fn`, which _creates_ a `Duration`.

You mention "complex calculations" but do not elaborate. Without an **actual** use case, which I've already indicated I do not have, I have no reason to add an API like this. You need to explain _why_ `Instant` isn't sufficient. Four bytes is very little, even on the smallest targets Rust supports.

gunnarpieter:
You and I both know that you edited `Duration::time_fn`. That just times a function. I have to do physics related calculations on  signals.

`Instant` is the most bare bones API that you can imagine. It does not even allow simple replacing of sub elements.

> You mention "complex calculations" but do not elaborate

I can't it has to do with signal processing. I recognize working with time is hard that is the sole reason I want to use your crate. But the inability to go from Offset to Primitive just makes it all the more awkward. 


If instants had a more elaborate API (similar to Primitive), I would use it. Is this something that you do feel ok with extending?

jhpratt:
> You and I both know that you edited Duration::time_fn.

Within one minute of posting because I got the method name wrong. I _never_ suggested using the entirety of `Duration`, nor do I appreciate the implication that I did. Edits are fully public, including the time they occurred.

> simple replacing of sub elements.

What's that?

> I can't it has to do with signal processing.

Then I'm afraid I can't help you in any way. The lack of a defined use case does not lead to things getting added. Declining to elaborate just means I still don't have a use case.

> If instants had a more elaborate API (similar to Primitive), I would use it. Is this something that you do feel ok with (allowing) extending?

`time::Instant` exists to permit some nicer APIs with `time::Duration`. Nothing else. It is otherwise identical to `std::time::Instant`. It is fundamentally opaque. These properties are all documented.

gunnarpieter:
lets just agree that this wont work out. I don't see the point in arguing with you. You and I are both set in our ways and it seems that there is little room for an agreement.

> The lack of a defined use case does not lead to things getting added. Declining to elaborate just means I still don't have a use case.

I never implied such a thing. I have accepted that it won't be added. If I was allowed to share more information I would.

> What's that?

Replacing of seconds, nanoseconds, etc.

I would like to add that there is something to say for providing access to the primitive.
Calling such a function would need to yield a primitive that has not been adjusted for different offsets.
As that is the whole merit of time Primitives that they are not offset depended or free from offsets and are solely based on UTC.
With the calls to Date and Time the whole premise of pureness of the primitive is discarded.

Anyways, Jacob, if you are still not inclined to add the functionality I would like to thank you for your time and work on this crate. Best wishes :)

jhpratt:
I'm not set in my ways, I just haven't seen a use case.

`Instant` doesn't have the concept of seconds and nanoseconds. It's an opaque value, even to me. Ultimately there is little I can do with the information I've been provided.

> As that is the whole merit of time Primitives that they are not offset depended or free from offsets and are solely based on UTC.

This is the fundamental point. Primitive means there is _no_ offset or an _unknown_ offset, not that the offset is UTC.

PR #468, opened 2022-05-05
Implement concise `Display` repr for `Duration`

### LINKED ISSUE #467: Alternative, concise `Display` format for `Duration`
While the current printouts for `Duration` are _precise_, they're not entirely pleasant to read:

```rust
println!("{}", time::Duration::new(123456, 789011223))
```

```
1d10h17m36s789ms11µs223ns
```

The lack of spacing probably doesn't help, but I think the main issue is that there's just too much information. 

For applications where conciseness is preferred, I suggest supporting the precision format specifier that prints only the largest nonzero field plus the given precision as a number of fractional digits, e.g.:

```rust
println!("{:.3}", time::Duration::new(123456, 789011223));
```

```
1.429d
```

abonander:
closes #467

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -748,38 +748,81 @@ impl Duration {
 // region: trait impls
 /// The format returned by this implementation is not stable and must not be relied upon.
 ///
+/// By default this produces an exact, full-precision printout of the duration.
+/// For a concise, rounded printout instead, you can use the `.N` format specifier:
+///
+/// ```
+/// # use time::Duration;
+/// #
+/// let duration = Duration::new(123456, 789011223);
+/// println!("{:.3}", duration);
+/// ```
+///
 /// For the purposes of this implementation, a day is exactly 24 hours and a minute is exactly 60
 /// seconds.
 impl fmt::Display for Duration {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        /// Format a single item.
-        macro_rules! item {
-            ($name:literal, $value:expr) => {
-                match $value {
-                    0 => Ok(()),
-                    value => value.fmt(f).and(f.write_str($name)),
-                }
-            };
-        }
-
-        if self.is_zero() {
-            return f.write_str("0s");
-        }
-
-        let seconds = self.seconds.unsigned_abs();
-        let nanoseconds = self.nanoseconds.unsigned_abs();
-
         if self.is_negative() {
             f.write_str("-")?;
         }
 
-        item!("d", seconds / 86_400)?;
-        item!("h", seconds / 3_600 % 24)?;
-        item!("m", seconds / 60 % 60)?;
-        item!("s", seconds % 60)?;
-        item!("ms", nanoseconds / 1_000_000)?;
-        item!("µs", nanoseconds / 1_000 % 1_000)?;
-        item!("ns", nanoseconds % 1_000)?;
+        if let Some(_precision) = f.precision() {
+            // Concise, rounded representation.
+
+            if self.is_zero() {
+                // Write a zero value with the requested precision.
+                return (0.).fmt(f).and_then(|_| f.write_str("s"));
+            }
+
+            /// Format the first item that produces a value greater than 1 and then break.
+            macro_rules! item {
+                ($name:literal, $value:expr) => {
+                    let value = $value;
+                    if value >= 1.0 {
+                        return value.fmt(f).and_then(|_| f.write_str($name));
+                    }
+                };
+            }
+
+            // Even if this produces a de-normal float, because we're rounding we don't really care.
+            let seconds = self.unsigned_abs().as_secs_f64();
+
+            item!("d", seconds / 86_400.);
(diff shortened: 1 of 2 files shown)


abonander:
It says it lost a line of coverage somewhere but the actual report shows the coverage didn't change?

src/duration.rs
```diff
@@ -742,38 +742,76 @@ impl Duration {
 // region: trait impls
 /// The format returned by this implementation is not stable and must not be relied upon.
 ///
+/// By default this produces an exact, verbose printout of the duration.
```
jhpratt, line 745: I think it would be better to describe the default as "precise", rather than "verbose".

src/duration.rs
```diff
@@ -742,38 +742,76 @@ impl Duration {
...
+/// For a concise, rounded printout instead, you can use the `.N` format specifier:
+///
+/// ```
+/// use time::Duration;
```
jhpratt, line 749: This should be hidden (by starting the line with `#`). This is because the import is implied — it's being shown in the `Duration` docs. The following blank line should either be removed or hidden as well.

src/duration.rs
```diff
@@ -742,38 +742,76 @@ impl Duration {
...
-            ($name:literal, $value:expr) => {
-                match $value {
-                    0 => Ok(()),
-                    value => value.fmt(f).and(f.write_str($name)),
-                }
-            };
-        }
-
         if self.is_zero() {
             return f.write_str("0s");
```
jhpratt, line 760: If the user is requesting a certain number of decimal places, we should always give them that many digits. I think this could be done by formatting `0f32` and appending the "s"`. It should still be "0s" by default, though.

src/duration.rs
```diff
@@ -742,38 +742,76 @@ impl Duration {
...
+                    if value >= 1.0 {
+                        return value.fmt(f).and_then(|_| f.write_str($name));
+                    }
+                };
+            }
+
+            // Even if this produces a de-normal float, because we're rounding we don't really care.
+            let seconds = self.abs().as_seconds_f64();
+
+            item!("d", seconds / 86_400.0);
```
jhpratt, line 783: I find floats to be more readable with the final "0" omitted. The decimal by itself is sufficient to tell the compiler it's a float.

jhpratt (changes requested):
Not sure where the lost line is either, but the coverage for the added code appears to be 100%, which is what matters.

src/duration.rs
abonander: "precise" would be redundant with "exact", I think. I'm just trying to get across the idea of "unless you want the exact duration down to the nanosecond, it's probably more verbose than you want".

src/duration.rs
abonander: I disagree. I think the trailing dot by itself looks out of place.

src/duration.rs
abonander: "exact, full-precsision" maybe?

jhpratt: "exact, full-precision" works for me.

src/duration.rs
jhpratt: It's already used in `src/ext.rs`; I'd like it to be consistent.

abonander:
@jhpratt I think this is ready for merge now.

jhpratt:
Ok, I'll take a look later. When I checked before CI was failing, but I see it's green now.

abonander:
Yeah I forgot to run the integration test locally but I fixed it.

src/duration.rs
```diff
@@ -742,38 +742,81 @@ impl Duration {
...
+                ($name:literal, $value:expr) => {
+                    let value = $value;
+                    if value >= 1.0 {
+                        return value.fmt(f).and_then(|_| f.write_str($name));
+                    }
+                };
+            }
+
+            // Even if this produces a de-normal float, because we're rounding we don't really care.
+            let seconds = self.abs().as_seconds_f64();
```
jhpratt, line 782: `self.abs()` saturates on overflow, which would provide an inaccurate output. Using `self.unsigned_abs()` should would. You may need to rebase, as the method was renamed to that a few days ago. The method returns `std::time::Duration`, so the method name to get the floating point value will be slightly different.

jhpratt (changes requested):
One edge case that needs consideration. Adding a test for this would probably be a good idea as well.

abonander:
@jhpratt fixed and rebased.

jhpratt:
For your sanity, I thought I'd post an update. Assuming you made the changes requested, this is good to go. It's just that I was quite busy before and am currently abroad (so on top of working I'm also sightseeing). I assure you I have not forgotten about this.

abonander:
^^ Maybe it'd be a good idea to turn off this bot until you're back then?

jhpratt:
Forgot to apply a label telling the bot to ignore it. My bad. My intent is to review it once ISO 8601 is complete. Though at this point I am considering doing this first.

jhpratt:
Implementation looks good. Sorry about the wait. Thank you!

PR #473, opened 2022-05-24
 Parse nulls as None on nested serde structs 

stusmall:
When parsing a structure nested inside an untagged enum serde treats
keys with a null value as Unit instead of calling visit_none.  This
causes an error of "invalid type: null, expected an
RFC3339-formatted `Option<OffsetDateTime>`".  This change just adds an
implementation for visit_unit so the value will be parsed as None as
expected.


jhpratt:
Would you mind rebasing? I think it's a matter of moving the new method into the newly-created macro.

jhpratt (approved):

PR #478, opened 2022-06-15
Fix incorrect RFC3339 offset parsing

zhu-he:
Fix a bug when parsing RFC3339 offset in range from `-00:01`  to `-00:59`.


jhpratt:
Good catch. Thank you!

PR #487, opened 2022-07-21
Add direct doc example for `datetime!` macro

xy2iii:
# Motivation

When using this crate, I wanted to create a PrimitiveDateTime. But I struggled to find how to create one. Eventually I landed on the datetime!() macro, but it didn't have an example usage. 

Now I know that `datetime!()` usage exemples are all over the doc, but  it's difficult to figure out the format from that, and there's no example in the exported macro tiself. This PR fixes that.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -30,6 +30,24 @@ pub use time_macros::date;
 ///
 /// [`OffsetDateTime`]: crate::OffsetDateTime
 /// [`PrimitiveDateTime`]: crate::PrimitiveDateTime
+///
+/// ```rust
+/// # use time::{Date, Month, macros::datetime, UtcOffset};
+/// assert_eq!(
+///     datetime!(2020-01-01 0:00),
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -30,6 +30,23 @@ pub use time_macros::date;
 ///
 /// [`OffsetDateTime`]: crate::OffsetDateTime
 /// [`PrimitiveDateTime`]: crate::PrimitiveDateTime
+///
+/// ```rust
+/// # use time::{Date, Time, Month, macros::datetime, PrimitiveDateTime, UtcOffset};
+/// assert_eq!(
+///     datetime!(2020-01-01 0:00),
+///     PrimitiveDateTime::new(Date::from_calendar_date(2020, Month::January, 1)?, Time::MIDNIGHT)
```
jhpratt, line 38:
```suggestion
///     Date::from_calendar_date(2020, Month::January, 1)?.midnight()
```

Same for the other two. Other than this nit, LGTM.

PR #489, opened 2022-07-22
Round subseconds to avoid float underflow

### LINKED ISSUE #488: ISO 8601 parsing from string sometimes is off by 1 nanosecond
Sometimes timestamps parsed from string are off by 1 nanosecond from the string text:
```rust
use time::{format_description::well_known::iso8601::Iso8601, macros::datetime, OffsetDateTime};

fn main() {
    let timestamp = "2022-07-22T12:52:50.349409+00:00";
    let parsed = OffsetDateTime::parse(timestamp, &Iso8601::DEFAULT).unwrap();
    // panic: `left` is `nanosecond: 349408999`, `right` is `nanosecond: 349409000`
    assert_eq!(parsed, datetime!(2022-07-22 12:52:50.349409000 UTC));
}
```

Better reproducer: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=31030fbca526a129ba26e1528bbe163a

mati865:
Fixes #488

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,6 @@ quickcheck-dep = { package = "quickcheck", version = "1.0.3", default-features =
 rand = { version = "0.8.4", optional = true, default-features = false }
 serde = { version = "1.0.126", optional = true, default-features = false }
 time-macros = { version = "=0.2.4", path = "time-macros", optional = true }
-num-traits = { version = "0.2.15", default-features = false }
 
 [target.'cfg(target_family = "unix")'.dependencies]
 libc = "0.2.98"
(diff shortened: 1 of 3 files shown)


mati865:
Round is unavailable on embedded, going to think how to solve it.

Cargo.toml
```diff
@@ -44,6 +44,7 @@ quickcheck-dep = { package = "quickcheck", version = "1.0.3", default-features =
 rand = { version = "0.8.4", optional = true, default-features = false }
 serde = { version = "1.0.126", optional = true, default-features = false }
 time-macros = { version = "=0.2.4", path = "time-macros", optional = true }
+num-traits = { version = "0.2.15", default-features = false }
```
jhpratt, line 47: Why not extract the one function required instead of adding a dependency? The hardware implementation can still be used, depending on the feature gate.

mati865: Sure, if you prefer it that way.

jhpratt:
The implementation will definitely be improved, but this is sufficient to merge. Thanks!

mati865:
Thanks!

PR #502, opened 2022-08-24
Add impl of arbitrary::Arbitrary for types

berkowski:
Adds implimentations of arbitrary::Arbitrary for the following:
  - Time
  - Duration
  - UtcOffset
  - PrimitiveDateTime
  - OffsetDateTime

Follows from the `quickcheck::Arbitrary` addition.  Otherwise projects using `time` that want to use `arbitrary` for fuzzing need to wrap everything in newtype structs and provide their own impls

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -85,7 +85,7 @@ jobs:
             --feature-powerset
             --optional-deps
             --group-features serde,rand
-            --exclude-features default,std,formatting,serde-human-readable,serde-well-known,local-offset,quickcheck,quickcheck-dep,time-macros,itoa,js-sys,wasm-bindgen
+            --exclude-features default,std,formatting,serde-human-readable,serde-well-known,local-offset,quickcheck,quickcheck-dep,time-macros,itoa,js-sys,wasm-bindgen,arbitrary,arbitrary-dep
             --features macros
             --exclude-all-features
             --target ${{ matrix.target.triple }}
@@ -106,7 +106,7 @@ jobs:
             --group-features serde,rand
             --group-features formatting,parsing
             --group-features serde-human-readable,serde-well-known
-            --exclude-features default,quickcheck-dep,time-macros,itoa,js-sys,wasm-bindgen
+            --exclude-features default,quickcheck-dep,time-macros,itoa,js-sys,wasm-bindgen,arbitrary,arbitrary-dep
             --features macros,local-offset
             --target ${{ matrix.target.triple }}
         if: matrix.target.has_std == true && matrix.target.has_local_offset == false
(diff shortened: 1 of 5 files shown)


jhpratt:
What advantage does this provide over the existing interoperability with `rand`? For `quickcheck` the advantage is shrinking, which does not appear to be the case with `arbitrary`.

berkowski:
The big difference seems to be that `arbitrary` uses an external fuzzer (they mention cargo-fuzz, which uses LLVM's libFuzzer, and afl.   Both are also part of the rust-fuzz github group) while `quickcheck` is an all-in-one solution that can be driven entirely with `cargo test`.  As such I expect those external fuzzers would handle the shrinking aspect, as the input used in `arbitrary` is simple a byte slice.

I don't have enough experience with fuzzing to comment on the pros and cons of either approach besides the obvious.  `arbitrary` does seem to have a decent amount of support, is used by 100+ crates given a quick glance at crates.io, and is actively maintained.  However to use it the types need to impl `arbitrary::Arbitrary`.  Without an impl in-crate one has to resort to newtype structs in each downstream crate.  I started down this path initially.  It was painful.

I came across `arbitrary` first before I found the `quickcheck` feature already in the crate.  For my particular use case either `arbitrary` or `quickcheck` will suffice as I actually just want to generate large sets of randomized data structures.  I figured since I had already started the pain of writing `arbitrary::Arbitrary` for some `time` structs I might as well finish it and submit a PR in case it saves someone else some effort down the road.

berkowski:
One other potential "feature" of `arbitrary` over `quickcheck` that comes to mind for my particular use case:  

I don't care about shrinking because I'm not actually fuzzing; I'm just generating lots of randomized data structures.  With `arbitrary` I have complete control over the random data being used to generate the structures, so I have the ability to have reproducible datasets if I use the same seed for the PRNG between runs.

I won't claim that it's extremely important for even my use, but I don't think that's possible with `quickcheck` as it treats `rand` explicitly as an internal implementation detail that users don't have access too.  That's not really a bad thing either as, well, you probably want randomness if you're actually fuzzing.

jhpratt:
Sorry, I was asking about why using `rand` would not suffice for anyone using `arbitrary`. I am aware of the differences between `quickcheck` and `arbitrary`. `rand` supports seeded RNGs, as it's ultimately just a trait being implemented.

berkowski:
They'd still have to implement `arbitrary::Arbitrary`, which would then require them to newtype all their `time`-based types.  Having the crate owner impl the trait solves that problem for all downstream users.

Beyond that I can't comment really on which approach to fuzzing is better.  Since I"m only interested in generating random structures then the `quickcheck` impl should work fine enough for my use case.

jhpratt:
After looking into this a little bit, I think what would be ideal is if `arbitrary` supported interoperability with any type supporting `rand`. It would (at first glance) take some relatively minor alterations to the internals of `arbitrary`, as it currently operates on `&[u8]`. If this were changed to `Cow<[u8]>`, it could implemented without too much effort.

I'm closing this PR because I think there is a significantly better alternative that would benefit the entire ecosystem, not just users of `time`.

Issue #505, opened 2022-09-06
Deserializing Weekday from u8 returns an error


Aegonek:
Trying to deserialize a Weekday from a number (without enabling `serde-human-readable` feature) returns an error `Error("invalid type: integer '1', expected a 'Weekday'", ...)`

I've created a minimal binary crate for testing, using stable channel. Dependency section:
```
[dependencies]
serde = "1.0.144"
time = { version = "0.3.14", features = [ "serde" ] }
serde_json = "1.0.85"
```

Code to reproduce the issue:
```
use std::error::Error;
use time::Weekday;

fn main() -> Result<(), Box<dyn Error>> {
    let day = Weekday::Monday;
    let str = serde_json::to_string(&day)?;
    let day: Weekday = serde_json::from_str(&str)?;
    println!("Day is {day}");
    Ok(())
}

// Result: Error: Error("invalid type: integer '1', expected a 'Weekday'", line: 1, column: 1)
```

I did some debugging and it seems that this code invokes `deserializer.deserialize_u8(Visitor::<Self>(PhantomData))` as expected, but deserialization still fails. I don't know Serde internals too well, so I don't know why. I could look for a solution later this week. You can assign me to this issue, please.

legas7:
Serde deserializer by default forwards `visit_u8/16/32` [to](https://github.com/serde-rs/serde/blob/master/serde/src/de/mod.rs#L1388) `visit_u64`. So weekday's (and month's) [visitors](https://github.com/time-rs/time/blob/main/src/serde/visitor.rs#L197) ` visit_u8` never gets called. One workaround could be changing visitor type to u64, but I'm wondering if there is more elegant way to override default forward, since we still serialized u8 value.

Here you can lookup what Deserialize and Visitor traits are about: https://serde.rs/impl-deserialize.html

Aegonek:
Hey, I had the issue figured out last week but no time to write. Using the example of Weekday to explain, but the implementation is faulty for other types too. Example code at the end. Marked interesting places with ^1 and ^2.

Issues:
1. ^1 - deserialize_u8 is only a hint about a type for a deserializer. Deserializer is free to ignore the hint and deserialize value as different type. JSON deserializer specifically decodes all numbers as i64. 

Proposed solution: none, none-issue, just not rely on this behavior.

2. ^2 - visit_u8 is only invoked if deserializer handles u8 specifically, which is kinda error-prone. Default behaviour for `visit_num` methods is:
[...]

PR #506, opened 2022-09-14
Fix error when deserializing datatypes from number

Aegonek:
https://github.com/time-rs/time/issues/505
Please see my comment in the issue.

### DIFF
--- a/src/serde/visitor.rs
+++ b/src/serde/visitor.rs
@@ -194,22 +194,6 @@ impl<'a> de::Visitor<'a> for Visitor<Weekday> {
         }
     }
 
-    fn visit_i64<E: de::Error>(self, value: i64) -> Result<Weekday, E> {
-        match value {
-            1 => Ok(Weekday::Monday),
-            2 => Ok(Weekday::Tuesday),
-            3 => Ok(Weekday::Wednesday),
-            4 => Ok(Weekday::Thursday),
-            5 => Ok(Weekday::Friday),
-            6 => Ok(Weekday::Saturday),
-            7 => Ok(Weekday::Sunday),
-            _ => Err(E::invalid_value(
-                de::Unexpected::Signed(value),
-                &"a value in the range 1..=7",
-            )),
-        }
-    }
-
     fn visit_u64<E: de::Error>(self, value: u64) -> Result<Weekday, E> {
         match value {
             1 => Ok(Weekday::Monday),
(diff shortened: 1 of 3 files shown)


src/serde/visitor.rs
```diff
@@ -204,7 +204,23 @@ impl<'a> de::Visitor<'a> for Visitor<Weekday> {
...
             7 => Ok(Weekday::Sunday),
             _ => Err(E::invalid_value(
-                de::Unexpected::Unsigned(value.into()),
+                de::Unexpected::Signed(value),
+                &"a value in the range 1..=7",
+            )),
+        }
+    }
+
+    fn visit_u64<E: de::Error>(self, value: u64) -> Result<Weekday, E> {
```
jhpratt, line 213: Both `visit_i64` and `visit_u64` are required to capture all integer types, from what I can tell? Excluding 128 bit types, but that's a separate story.

tests/integration/serde/json.rs
```diff
@@ -0,0 +1,144 @@
...
+    assert_eq!(serialize(March, Readable)?, "\"March\"");
+    assert_eq!(deserialize::<Month>("\"March\"", Readable)?, March);
+    assert_eq!(deserialize::<Month>("3", Readable)?, March);
+
+    Ok(())
+}
+
+#[test]
+fn time_json() -> Result<(), Box<dyn Error>> {
+    let time = Time::from_hms(12, 40, 20)?;
```
jhpratt, line 77: For statically known values, it's best to use the macro.

jhpratt (changes requested):
Couple changes. CI should also pass — right now there seems to be quite a bit uncovered. [Here's a link](https://app.codecov.io/gh/time-rs/time/pull/506) to the coverage for this PR.

src/serde/visitor.rs
Aegonek: Yes, it seems to be the case. In this commit I added `visit_u64` implementation for every `visit_i64` implementation.

tests/integration/serde/json.rs
Aegonek: :+1:

Aegonek:
> Couple changes. CI should also pass — right now there seems to be quite a bit uncovered. [Here's a link](https://app.codecov.io/gh/time-rs/time/pull/506) to the coverage for this PR.

Yeah, Codecov says that `visit_i64` is not tested. It seems that `serde_json` will always delegate `deserialize_u8` to `visit_u64` for our use case. It may delegate it to `deserialize_i64` if the number is negative and `deserialize_f64` if the number is greater than `u64::MAX` - but we know that won't happen. I think the best course of action is ^1: to remove `visit_i64` implementation.

Alternative (^2) - add implementation for `visit_f64`, opt out from code coverage for `visit_i64` and `visit_f64`? Given that those methods exist mainly for "better safe than sorry", I think that would also be fine. Although keeping methods you don't expect to be used by framework is a bit smelly, so I prefer option ^1.

jhpratt (approved):
I just pushed up a small diff to simplify some of the code, which reduces its size slightly. LGTM at this point. Thank you!

PR #516, opened 2022-10-25
Only include libc and num_threads if compiling on unix with local offset

MarcusGrass:
libc and num_threads are only used on linux with the `local_offset` feature so they can be made optional and only included then. 
At least it worked compiling for x86_64-pc-windows-gnu without errors and the dependencies were not included.


time/Cargo.toml
```diff
@@ -26,7 +26,7 @@ default = ["std"]
 alloc = ["serde?/alloc"]
 formatting = ["dep:itoa", "std", "time-macros?/formatting"]
 large-dates = ["time-macros?/large-dates"]
-local-offset = ["std"]
+local-offset = ["std", "libc", "num_threads"]
```
jhpratt, line 29:
```suggestion
local-offset = ["std", "dep:libc", "dep:num_threads"]
```

This avoids introducing implicit feature flags.

time/src/macros.rs
```diff
@@ -65,7 +65,6 @@ pub use time_macros::datetime;
 /// );
 /// # Ok::<_, time::Error>(())
 /// ```
-///
```
jhpratt, line 68: You can remove these changes. I know `rustfmt` is failing, but it's a bug upstream and not something that needs to be changed here.

jhpratt (changes requested):
I would've sworn that this wasn't supported by Cargo, but it definitely works when I have this locally. So I see no reason to not do this! Just a minor change.

time/Cargo.toml
MarcusGrass: Yeah there's so much stuff you can do that I don't know about, the dep thing is nice though, should have spotted it from the other imports!

jhpratt: No worries! It's relatively new, only about 7 months old on stable. I did add a note (mostly for myself) above `[dependencies]`, but you had no reason to look there.

jhpratt:
Thanks again!

PR #524, opened 2022-11-18
New Ranges Feature

ferdinandkeller:
I've started to implement proper ranges for the library. This pull request is not one that is ready to be merged with the main branch. Many functions don't have an implementation yet, and it still lacks some features I would like to add. But I would like some feedback before progressing in the direction I am going.

# What is new

Here is a quick summary of what is present:
- Extention of the `OffsetDateTime` struct, adding functions likes `floor_seconds()`, `ceil_seconds()` and `next_second()`. They are necessary to the `OffsetDateTimeRangeExt`.
- Extension of the `Range<OffsetDateTime>` struct with functions like `days()`, `full_days()`, `overlapping_days()`, `intersection()`, `union()`, ...
- Extension of `&[Range<OffsetDateTime>]` slices (no functions implemented yet).

Important notes about ranges:
- You need two `OffsetDateTime` to create one range, so it was decided that the ranges will always work with the offset of the start datetime.
- If the ranges are reversed (end < start), most methods on this range will return `None` or an empty `Vec`.

# Questions I might have

Should ranges implement the `Copy` trait ? The struct `Range<OffsetDateTime>` never changes, and doesn't implement the `Iterator` trait, so there shouldn't be any problem with this. I would like your feedback on the idea.

EDIT: You can't implement the `Copy` trait, as the `Range` struct is not defined inside the crate.

# Coming improvements

- Implementing missing functions
- Implementing `OffsetDateTimeExt` for `RangeInclusive`
- Parsing & Formatting for ranges
- More unit tests
- Double sided iterators
- Range creation macro ?
- Similar functions for millis, micros and nanos ?

### DIFF
--- a/tests/main.rs
+++ b/tests/main.rs
@@ -137,6 +137,7 @@ require_all_features! {
     mod macros;
     mod meta;
     mod month;
+    mod offset_date_time_range;
     mod offset_date_time;
     mod parse_format_description;
     mod parsed;
--- a/tests/offset_date_time.rs
+++ b/tests/offset_date_time.rs
@@ -529,6 +529,319 @@ fn replace_nanosecond() -> Result<()> {
 }
 
 #[test]
+fn floor_seconds() {
+    assert_eq!(
+        datetime!(2022-04-18 03:17:12.123_456_789 UTC).floor_seconds(),
+        datetime!(2022-04-18 03:17:12 UTC)
+    );
+    assert_eq!(
+        datetime!(2022-04-18 03:17:12 UTC).floor_seconds(),
+        datetime!(2022-04-18 03:17:12 UTC)
+    );
+}
+
+#[test]
+fn floor_minutes() {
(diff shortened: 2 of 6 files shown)


jhpratt:
Just a heads up, it'll probably be a few days before I get to looking at this.

jhpratt:
Yeah, when I get around to it (possibly this coming weekend), I'll be looking at the public API, not the implementation.

jhpratt:
After a quick review of the public API, I'm honestly impressed that you managed to get this working — it's a complicated thing to manage. With that said, I am (unfortunately) closing this pull request, as I feel it's far too large for me to manage at this point in time. I am quite busy working on language-related matters, so any issues that pop up would require me to spend time that I don't have.

Thank you for your contribution regardless! If/when I come around to having iterator helpers, I will definitely look at this code for guidance.

PR #526, opened 2022-11-18
Remove unnecessary floating point remainder

CryZe:
Because the code handles all the edge cases already, we can simply cast the integer seconds back to a floating point number to truncate it, instead of using `seconds % 1.0` which is not only slower, but also bloats the binary, especially on `no_std`, as it brings in an implementation of `fmod` / `fmodf`.

### DIFF
--- a/time/src/duration.rs
+++ b/time/src/duration.rs
@@ -326,7 +326,10 @@ impl Duration {
         if seconds.is_nan() {
             crate::expect_failed("passed NaN to `time::Duration::seconds_f64`");
         }
-        Self::new_unchecked(seconds as _, ((seconds % 1.) * 1_000_000_000.) as _)
+        let seconds_truncated = seconds as i64;
+        // This only works because we handle the overflow condition above.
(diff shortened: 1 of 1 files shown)


jhpratt:
Neat optimization. I've confirmed that it reduces the amount of generated assembly, so it's definitely a win.

CryZe:
Another thing to consider might be to port the algorithm from std, which nowadays basically abuses the fact that floating point numbers are really just the mantissa that represents "two integers" separated by a decimal point (through the exponent). So you could just extract those two integers without any loss of precision (as would be the case when trying to represent the fraction as its own floating point number).

jhpratt:
Ah right, I forgot the algorithm changed a couple releases ago. Thanks for the reminder!

jhpratt:
So…I just attempted to adapt that algorithm. It's a bit hacky, as I basically just forced the sign to be what it was supposed to be at the end, and still doesn't differentiate overflow and NaN (for the panic message). But in the end I am measuring a _regression_. That may be because of the exact benchmarks I am using, but it doesn't seem to be a super obvious gain. As such I'm going to keep your improvement and stop there.

PR #527, opened 2022-11-18
Fix Compilation on WebAssembly

CryZe:
Apparently the latest version of `main` doesn't compile to WebAssembly if you don't activate the `wasm-bindgen` feature. Also nowadays the way to check for WebAssembly is different, so I also changed that across the board.


jhpratt:
Thanks!

PR #529, opened 2022-11-25
Add millis functions and serde for unix_timestamp

kujeger:
Hi!

First off, thanks a lot for this excellent crate.

I've been happily using `time` for a few projects, but often have the need to deserialize and parse data featuring unix timestamps in millisecond format -- usually created in some java or kotlin application where this seems to be very common.

Instead of only doing local altering/parsing of this, it seems like it could be a good fit along with the existning nanosecond support.

I've tried to make the new code as close as possible to the existing nanosecond/second functions where possible.

### DIFF
--- a/tests/offset_date_time.rs
+++ b/tests/offset_date_time.rs
@@ -76,6 +76,19 @@ fn from_unix_timestamp() {
 }
 
 #[test]
+fn from_unix_timestamp_millis() {
+    assert_eq!(
+        OffsetDateTime::from_unix_timestamp_millis(0),
+        Ok(OffsetDateTime::UNIX_EPOCH),
+    );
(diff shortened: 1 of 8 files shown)


jhpratt:
Ultimately I want to add a `[unix_timestamp]` component, which could have support for modifiers to determine if it's second, millisecond, microsecond, or nanosecond. As such I would prefer not to add the serde helpers here.

As to the non-serde methods, there is no reason someone cannot simply multiple the value by 1,000,000 (for milliseconds) or 1,000 (for microseconds) and then pass the value to `from_unix_timestamp_nanos`. For getting the value, it can be divided by the same amount. For this reason I do not see the need to add these methods at the present time.

Issue #542, opened 2023-01-26
Weekday from number


Kinrany:
Currently weekdays can be converted to numbers, with four options: zero-based/one-based and sunday-based/monday-based.

It would be convenient to have the reverse constructors: from a zero-based or one-based offset, for a sunday-based or monday-based week, create a Weekday value.

jhpratt:
This is definitely doable with minimal effort. Once I get to documenting features that are on `main` and cut a release, I can do this quite quickly.

Kinrany:
I can probably help with a PR. Unless it's easier for you to just do it compared to reviewing code later.

jhpratt:
Up to you! Either way it will be in a separate release.

PR #544, opened 2023-01-26
Implement `Weekday::nth_next`

Kinrany:
Adds `Weekday::nth`, a `fn(Weekday, u8) -> Weekday` method.

This is similar to `Iterator::nth`: it returns the n-th next week day after the current one.

This is in place of four different methods with long names that would be opposites of `number_from_monday`, `number_from_sunday`, `num_days_from_monday`, `num_days_from_sunday`. I was bikeshedding the names in my mind and realized that making a single general method is likely more ergonomic, if only because of the amount of effort it was taking to write - and therefore understand - the fairly long descriptions of each method.

Closes https://github.com/time-rs/time/issues/542

### DIFF
--- a/benchmarks/weekday.rs
+++ b/benchmarks/weekday.rs
@@ -25,26 +25,18 @@ setup_benchmark! {
     }
 
     fn nth(ben: &mut Bencher<'_>) {
-        ben.iter(|| Sunday.nth(0));
-        ben.iter(|| Sunday.nth(1));
-        ben.iter(|| Sunday.nth(2));
-        ben.iter(|| Sunday.nth(3));
-        ben.iter(|| Sunday.nth(4));
-        ben.iter(|| Sunday.nth(5));
-        ben.iter(|| Sunday.nth(6));
+        ben.iter(|| Sunday.nth_next(0));
+        ben.iter(|| Sunday.nth_next(1));
+        ben.iter(|| Sunday.nth_next(2));
+        ben.iter(|| Sunday.nth_next(3));
+        ben.iter(|| Sunday.nth_next(4));
+        ben.iter(|| Sunday.nth_next(5));
+        ben.iter(|| Sunday.nth_next(6));
 
-        ben.iter(|| Monday.nth(0));
-        ben.iter(|| Monday.nth(1));
-        ben.iter(|| Monday.nth(2));
-        ben.iter(|| Monday.nth(3));
-        ben.iter(|| Monday.nth(4));
-        ben.iter(|| Monday.nth(5));
-        ben.iter(|| Monday.nth(6));
-
-        ben.iter(|| Sunday.nth(7));
-        ben.iter(|| Sunday.nth(u8::MAX));
-        ben.iter(|| Monday.nth(7));
-        ben.iter(|| Monday.nth(u8::MAX));
+        ben.iter(|| Sunday.nth_next(7));
+        ben.iter(|| Sunday.nth_next(u8::MAX));
+        ben.iter(|| Monday.nth_next(7));
+        ben.iter(|| Monday.nth_next(u8::MAX));
     }
 
     fn number_from_monday(ben: &mut Bencher<'_>) {
--- a/benchmarks/weekday.rs
+++ b/benchmarks/weekday.rs
@@ -24,6 +24,29 @@ setup_benchmark! {
         ben.iter(|| Saturday.next());
     }
 
+    fn nth(ben: &mut Bencher<'_>) {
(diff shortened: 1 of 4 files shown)


Kinrany:
I don't think I know what to do with those seemingly unrelated Clippy errors 😓

jhpratt:
Don't worry about clippy. That's a side effect of there being a Rust release yesterday that includes new lints. However, code coverage is also failing, so presumably some more tests are needed.

Kinrany:
Hmm, it's asking to cover an unreachable case. Wish there was a way to construct the enum from a raw u8 without bringing a whole extra package.

jhpratt:
Alright. I'll have a concrete suggestion when I get to reviewing the code.

benchmarks/weekday.rs
```diff
@@ -24,6 +24,29 @@ setup_benchmark! {
...
+        ben.iter(|| Sunday.nth(5));
+        ben.iter(|| Sunday.nth(6));
+
+        ben.iter(|| Monday.nth(0));
+        ben.iter(|| Monday.nth(1));
+        ben.iter(|| Monday.nth(2));
+        ben.iter(|| Monday.nth(3));
+        ben.iter(|| Monday.nth(4));
+        ben.iter(|| Monday.nth(5));
+        ben.iter(|| Monday.nth(6));
```
jhpratt, line 42: Any particular reason for duplicating this from `Sunday`?

tests/weekday.rs
```diff
@@ -1,3 +1,4 @@
+use quickcheck_macros::quickcheck;
```
jhpratt, line 1: All existing quickcheck tests are in the `tests/quickcheck.rs` file.

time/src/weekday.rs
```diff
@@ -66,6 +66,26 @@ impl Weekday {
...
     }
 
+    /// Get n-th next day.
+    ///
+    /// ```rust
+    /// # use time::Weekday;
+    /// assert_eq!(Weekday::Monday.nth(1), Weekday::Tuesday);
+    /// assert_eq!(Weekday::Sunday.nth(10), Weekday::Wednesday);
+    /// ```
+    pub const fn nth(self, n: u8) -> Self {
```
jhpratt, line 76:
```suggestion
    pub const fn nth_next(self, n: u8) -> Self {
```

time/src/weekday.rs
```diff
@@ -66,6 +66,26 @@ impl Weekday {
...
+        match (self.number_days_from_monday() + (n % 7)) % 7 {
+            0 => Monday,
+            1 => Tuesday,
+            2 => Wednesday,
+            3 => Thursday,
+            4 => Friday,
+            5 => Saturday,
+            6 => Sunday,
+            _ => unreachable!(),
```
jhpratt, line 85:
```suggestion
            val => {
                debug_assert_eq!(val, 6);
                Sunday
            }
```

This will ensure complete coverage.

time/src/weekday.rs
```diff
@@ -66,6 +66,26 @@ impl Weekday {
...
 
+    /// Get n-th next day.
+    ///
+    /// ```rust
+    /// # use time::Weekday;
+    /// assert_eq!(Weekday::Monday.nth(1), Weekday::Tuesday);
+    /// assert_eq!(Weekday::Sunday.nth(10), Weekday::Wednesday);
+    /// ```
+    pub const fn nth(self, n: u8) -> Self {
+        match (self.number_days_from_monday() + (n % 7)) % 7 {
```
jhpratt, line 77:
```suggestion
        match (self.number_days_from_monday() + n % 7) % 7 {
```

Operator precedence means we don't need parentheses here.

jhpratt (changes requested):

jhpratt:
@Kinrany Would you mind addressing the requested changes above? Otherwise I'll likely end up implementing the request myself.

Kinrany:
Sorry @jhpratt. Other things keep taking priority. I should be able to find a moment to do it eventually, but I wouldn't mind at all if you did reimplement it. I don't care about attribution, this is just something I wanted to have in the library.

Thank you for the hard work of maintaining the package, you rule 😄

benchmarks/weekday.rs
Kinrany: I wanted to have tests for values stored internally as both 0 and non-0.

jhpratt:
Up to you! I don't have a release planned for anything else at the moment, so it's not like there's a big rush.

Kinrany:
Thank you!

PR #545, opened 2023-01-29
Fix new Rust 1.67 clippy errors

Kinrany:
Fixes new Clippy warnings that came with Rust 1.67.

Warnings:

 * [uninlined_format_args](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args)
 * [unchecked_duration_subtraction](https://rust-lang.github.io/rust-clippy/master/index.html#unchecked_duration_subtraction)

Also fixes a formatting issue: comments around [undocumented_unsafe_blocks](https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks): `//false positive` was being considered part of the same comment block with the following `// SAFETY: ...` comment, so the latter was indented to be on the same level. Like

```rust
let x = 0; // first value
           // will be used later
```

Rust-analyzer and cargo fmt also seemingly disagreed on the correct way to format [time/src/formatting/mod.rs#L128](https://github.com/time-rs/time/blob/main/time/src/formatting/mod.rs#L128).

### DIFF
--- a/time/src/date_time.rs
+++ b/time/src/date_time.rs
@@ -93,9 +93,10 @@ impl MaybeOffset for offset_kind::None {
     type OffsetType = ();
     type OffsetKind = Self;
 
-    #[allow(clippy::undocumented_unsafe_blocks)] // false positive
-    // SAFETY: `()` is not `UtcOffset`.
-    const HAS_OFFSET: UnsafeBool = unsafe { UnsafeBool::new(false) };
+    const HAS_OFFSET: UnsafeBool = {
+        // SAFETY: `()` is not `UtcOffset`.
+        unsafe { UnsafeBool::new(false) }
+    };
 
     fn as_offset(_: ()) -> Option<UtcOffset> {
         None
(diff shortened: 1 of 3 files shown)


time/src/instant.rs
```diff
@@ -190,11 +195,16 @@ impl Add<StdDuration> for Instant {
 impl_add_assign!(Instant: Duration, StdDuration);
 impl_add_assign!(StdInstant: Duration);
 
+/// # Panics
```
jhpratt, line 198: This documentation should be on the method, not the trait.

time/src/date_time.rs
```diff
@@ -114,9 +115,10 @@ impl MaybeOffset for offset_kind::Fixed {
     type OffsetType = UtcOffset;
     type OffsetKind = Self;
 
-    #[allow(clippy::undocumented_unsafe_blocks)] // false positive
-    // Safety: `UtcOffset` is the offset type.
-    const HAS_OFFSET: UnsafeBool = unsafe { UnsafeBool::new(true) };
+    const HAS_OFFSET: UnsafeBool = {
+        // Safety: `UtcOffset` is the offset type.
+        unsafe { UnsafeBool::new(true) }
+    };
```
jhpratt, line 121: I don't care for this change. The existing code is acceptable, and it's just a bug with clippy.

jhpratt (changes requested):
This does more than just fix clippy errors. But regardless, some changes are requested. There is also a warning in the tests that was not addressed. I'm not sure what formatting issue you're referring to — formatting is checked in CI on every push.

jhpratt:
I am making the changes myself so that I can move ahead with other work without CI failing. Closing for that reason.

Kinrany:
Absolutely, sorry I didn't have time! I'm glad if the PR was of any help.

jhpratt:
Not an issue! I ended up pushing changes similar to what would have been the end result of this PR.

Issue #327, opened 2021-05-26
The Book is hardly readable


qiujiangkun:
This book https://time-rs.github.io/book has almost nothing in it. Is it complete?

jhpratt:
No, it's not.

SorteKanin:
It would be helpful if the book made it clear that it is incomplete - right now the missing text seems like a bug :)

briankung:
This put me off using this crate for several days trying to find the actual book.

jhpratt:
Incidentally I was just thinking about this last night. I think it may be preferable to inline this into crate documentation, using empty modules for hierarchy. This would make it more discoverable and more likely to be maintained once it's actually written.

If you're looking for somewhere to start, though, documenting formatting and parsing is probably the best place initially. Other than that I don't think there's a particular order that anything should be completed in.

This is, quite obviously, a low priority for me. There's development of the time crate itself, a number of RFCs that I am working on for Rust itself, and other improvements in Rust that are where I've been putting my focus. I certainly appreciate the offer to help! I assure you I will review any pull requests that are sent my way, even if they may be delayed (like the couple open now).

laralove143:
i also think there should at least be an incomplete warning in the main page of the book and in the website https://time-rs.github.io

egrieco:
There's a reason the US Army puts "This page intentionally left blank." in their various manuals. The book almost scared me away from this crate as well.

jhpratt:
Everyone: I'm super busy. The website is open source. Feel free to create a pull request with the notice. Or better yet, help write some documentation :wink: Repeated comments here don't do anything.

egrieco:
Fully understand @jhpratt. Been working on other code but I'll try to get to this in the not too distant future.

I would feel a little bit bad just adding a whole bunch of "Intentionally left blank." comments. However, now that I'm actually using the library, there are a few other code changes that might up the priority of creating a pull request.

xy2iii:
Am championning this. @jhpratt anything to keep in mind?

jhpratt:
Not really.

sourcefrog:
(Let me first say I absolutely understand this is maintained on a volunteer basis and this comment is posted with an intention to be useful rather than to complain.)

I also just looked at <https://docs.rs/time/0.3.12/time/index.html> which has no introduction to the crate but rather just begins with a list of feature flags. It makes the crate look less mature than it apparently is. 

[...]

PR #554, opened 2023-02-27
Add doc links to main doc and formatting

nw0:
Hi, occasional user of this crate here! I've seen #327 and found the book reasonably helpful, but it was not at all obvious to me that the book existed at first.

This PR adds a link to the book at the top of the crate's rustdoc, as well as some intra-doc links for the formatting items.

### DIFF
--- a/time/src/formatting/formattable.rs
+++ b/time/src/formatting/formattable.rs
@@ -13,7 +13,7 @@ use crate::{error, Date, Time, UtcOffset};
 
 /// A type that describes a format.
 ///
-/// Implementors of [Formattable] are [format descriptions](crate::format_description).
+/// Implementors of [`Formattable]` are [format descriptions](crate::format_description).
 ///
 /// [Date::format] and [Time::format] each use a format description to generate
 /// a String from their data. See the respective methods for usage examples.
--- a/time/src/formatting/formattable.rs
+++ b/time/src/formatting/formattable.rs
@@ -13,9 +13,9 @@ use crate::{error, Date, Time, UtcOffset};
 
 /// A type that describes a format.
 ///
-/// Implementors of [`Formattable]` are [format descriptions](crate::format_description).
+/// Implementors of [`Formattable`] are [format descriptions](crate::format_description).
 ///
-/// [Date::format] and [Time::format] each use a format description to generate
+/// [`Date::format`] and [`Time::format`] each use a format description to generate
 /// a String from their data. See the respective methods for usage examples.
 #[cfg_attr(__time_03_docs, doc(notable_trait))]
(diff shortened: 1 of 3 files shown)


time/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+//! A safe date and time library for Rust.
+//!
+//! [The book] is the main documentation for `time`.
```
jhpratt, line 3: This definitely isn't the case. Right now the only thing it's the primary reference for is format descriptions. Most of the book is still empty, as I've focused my efforts elsewhere.

time/src/formatting/formattable.rs
```diff
@@ -11,7 +11,12 @@ use crate::formatting::{
 };
 use crate::{error, Date, Time, UtcOffset};
 
-/// A type that can be formatted.
+/// A type that describes a format.
+///
+/// Implementors of [Formattable] are [format descriptions](crate::format_description).
```
jhpratt, line 16:
```suggestion
/// Implementers of [`Formattable]` are [format descriptions](crate::format_description).
```

Spelling and putting types/traits in inline code blocks. The latter should be done in other places as well.

jhpratt (changes requested):

time/src/lib.rs
nw0: Would you prefer a different text here, or perhaps a badge? I do recall struggling with format descriptions before finding the book.

time/src/formatting/formattable.rs
nw0: Rustdoc generates "Implementors" with an 'o': e.g. <https://docs.rs/time/latest/time/formatting/trait.Formattable.html#implementors> and it is the spelling used in the language documentation.

Will update the code blocks.

time/src/lib.rs
jhpratt: In that specific location, I think it could go without until there's more content. However, adding a reference to the relevant pages of the book on the various format/parse methods would make sense.

time/src/formatting/formattable.rs
jhpratt: Fair enough on the spelling, even if that's not how I would spell it.

time/src/lib.rs
nw0: I'll remove this change for this PR, the format/parse methods are reasonably documented IMO.

time/src/formatting/formattable.rs
jhpratt: Heh, looks like you blindly accepted the suggested change I had. I mistyped it — the trailing backtick is in the wrong location. That will need to be fixed, with the other locations also receiving backticks as appropriate. CI should also pass.

PR #556, opened 2023-02-28
Update `serde::format_description!()` docs

TristanDebrunner:
Recent updates to the `serde::format_description!()` macro allow format descriptions other than format strings to be passed. This updates the documentation to reflect that.

### DIFF
--- a/time/src/format_description/well_known/iso8601/adt_hack.rs
+++ b/time/src/format_description/well_known/iso8601/adt_hack.rs
@@ -11,7 +11,7 @@ use super::{Config, DateKind, FormattedComponents as FC, OffsetPrecision, TimePr
 #[doc(hidden)]
 pub type DoNotRelyOnWhatThisIs = u128;
 
-/// An encoded [`Config`] that can be used as a const parameter to [`Iso8601`].
+/// An encoded [`Config`] that can be used as a const parameter to [`Iso8601`](super::Iso8601).
 ///
 /// The type this is aliased to must not be relied upon. It can change in any release without
 /// notice.
@@ -49,10 +49,11 @@ impl<const CONFIG: EncodedConfig> Iso8601<CONFIG> {
 }
 
 impl Config {
-    /// Encode the configuration, permitting it to be used as a const parameter of [`Iso8601`].
+    /// Encode the configuration, permitting it to be used as a const parameter of
+    /// [`Iso8601`](super::Iso8601).
     ///
-    /// The value returned by this method must only be used as a const parameter to [`Iso8601`]. Any
-    /// other usage is unspecified behavior.
+    /// The value returned by this method must only be used as a const parameter to
+    /// [`Iso8601`](super::Iso8601). Any other usage is unspecified behavior.
     pub const fn encode(&self) -> EncodedConfig {
         let mut bytes = [0; EncodedConfig::BITS as usize / 8];
 
--- a/time/src/serde/mod.rs
+++ b/time/src/serde/mod.rs
@@ -27,23 +27,39 @@ use core::marker::PhantomData;
 #[cfg(feature = "serde-human-readable")]
 use serde::ser::Error as _;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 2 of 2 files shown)


time/src/serde/mod.rs
```diff
@@ -27,24 +27,34 @@ use core::marker::PhantomData;
...
+/// Invoked as `serde::format_description!(mod_name, Date, FORMAT)` where `FORMAT` is either a
+/// `"<format string>"` or something that implements
+#[cfg_attr(
+    feature = "formatting",
+    doc = "[`Formattable`](crate::formatting::Formattable)"
+)]
+#[cfg_attr(not(feature = "formatting"), doc = "`Formattable`")]
+/// and/or
+#[cfg_attr(feature = "parsing", doc = "[`Parsable`](crate::parsing::Parsable)")]
+#[cfg_attr(not(feature = "parsing"), doc = "`Parsable`")]
```
jhpratt, line 46: It would be better to only show the traits that need to be implemented, rather than showing a trait that isn't available with the current feature set.

time/src/serde/mod.rs
```diff
@@ -27,24 +27,34 @@ use core::marker::PhantomData;
...
+/// modules are only visible in the current scope.
 ///
 /// The returned `Option` will contain a deserialized value if present and `None` if the field
 /// is present but the value is `null` (or the equivalent in other formats). To return `None`
 /// when the field is not present, you should use `#[serde(default)]` on the field.
 ///
 /// # Examples
 ///
+/// Using a format string:
 /// ```rust,no_run
```
jhpratt, line 58: Code blocks should have a blank line before them.

jhpratt (changes requested):
Overall, looks good! There are a few minor changes requested, in addition to making sure CI passes.

time/src/format_description/well_known/iso8601/adt_hack.rs
```diff
@@ -11,7 +11,7 @@ use super::{Config, DateKind, FormattedComponents as FC, OffsetPrecision, TimePr
 #[doc(hidden)]
 pub type DoNotRelyOnWhatThisIs = u128;
 
-/// An encoded [`Config`] that can be used as a const parameter to [`Iso8601`].
+/// An encoded [`Config`] that can be used as a const parameter to [`Iso8601`](super::Iso8601).
```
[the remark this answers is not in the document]
TristanDebrunner: With all features enabled, that is true, but without this change, `cargo doc --features=macros,parsing,serde` failed for me:
```
error: unresolved link to `Iso8601`
  --> time/src/format_description/well_known/iso8601/adt_hack.rs:14:70
   |
14 | /// An encoded [`Config`] that can be used as a const parameter to [`Iso8601`].
   |                                                                      ^^^^^^^ no item named `Iso8601` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
```

(this is with cargo 1.66.1)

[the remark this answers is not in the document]
TristanDebrunner: The import that the link relies on is gated on `formatting` (lines 5&6):
```
#[cfg(feature = "formatting")]
use super::Iso8601;
```

time/src/serde/mod.rs
TristanDebrunner: `rustfmt` is removing those lines

TristanDebrunner: Actually, it's only sometimes doing it, need to look into it more...

TristanDebrunner: Yeah, `rustfmt` is removing those empty lines on the second and third examples, but not the first. I haven't been able to figure out why, nor have I found a way to circumvent it...

jhpratt:
I added a minor commit fixing formatting (due to the way HTML handles whitespace) and correcting a typo. I also confirmed that rustfmt doesn't handle the blank lines the way I wanted — I suspect this is due to the use of `#[cfg(foo, doc = "")]`, which doesn't interact well with formatting.

Thank you!

PR #561, opened 2023-03-28
add support for bincode v2

irevoire:
Hey!
I don’t know if this is the kind of PR you would be willing to merge, but I would really like to use bincode in our workflow, and the only blocking crate is time, so I was wondering if it would be possible to add the support of bincode under a feature flag.

### DIFF
--- a/time/Cargo.toml
+++ b/time/Cargo.toml
@@ -31,6 +31,7 @@ macros = ["dep:time-macros"]
 parsing = ["time-macros?/parsing"]
 quickcheck = ["dep:quickcheck", "alloc"]
 rand = ["dep:rand"]
+bincode = ["dep:bincode"]
 serde = ["dep:serde", "time-macros?/serde"]
 serde-human-readable = ["serde", "formatting", "parsing"]
 # Deprecated in favor of using the relevant flags directly.
@@ -44,6 +45,7 @@ wasm-bindgen = ["dep:js-sys"]
 itoa = { version = "1.0.1", optional = true }
 quickcheck = { version = "1.0.3", default-features = false, optional = true }
 rand = { version = "0.8.4", optional = true, default-features = false }
+bincode = { version = "2.0.0-rc", default-features = false, optional = true }
 serde = { version = "1.0.126", optional = true, default-features = false }
 time-core = { version = "=0.1.0", path = "../time-core" }
 time-macros = { version = "=0.2.8", path = "../time-macros", optional = true }
--- a/time/src/date.rs
+++ b/time/src/date.rs
@@ -32,6 +32,7 @@ pub(crate) const MAX_YEAR: i32 = if cfg!(feature = "large-dates") {
 /// inclusive by enabling the `large-dates` crate feature. Doing so has performance implications
 /// and introduces some ambiguities when parsing.
 #[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[cfg_attr(bincode, derive(bincode::Encode, bincode::Decode))]
 pub struct Date {
     /// Bitpacked field containing both the year and ordinal.
     // |     xx     | xxxxxxxxxxxxxxxxxxxxx | xxxxxxxxx |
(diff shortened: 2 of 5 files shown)


irevoire:
Hey,

> What advantage does this give over using the native support for any type implementing Serialize/Deserialize?

So here are the two points of why I want to use bincode;
- bincode v2 is generally faster than bincode+serde ([with lto enabled](https://github.com/bincode-org/bincode/issues/618))
- A lot of attributes you can use in serde breaks bincode+serde, while it always works with Encode/Decode

> Also, does the derive not implicitly guarantee a stable ABI on my end?

That’s something else I wanted to talk about with you before merging anything. I _think_ what I did is the worst way to implement Encode/Decode. tbh I did it because I needed it fast, but yes, it forces you to never change your structures (at least in a patch version), + it’s not optimal at all, I think.

I don’t know much about the crate and dates in general, so let me know if I’m wrong, but I think we would be better off implementing our own Encode/Decode. Do you think storing a single `unix_timestamp()` of the `OffsetDateTime` would be enough? Should we also encode the offset?

irevoire:
> Is this a technical restriction or merely unimplemented? If the latter, it's something that should be fixed upstream.

That's a technical restriction, it can’t be implemented, and that’s the reason for all this bincodev2 thingy

jhpratt:
I'd likely need more of an explanation of the actual restriction if this were to be merged. However, the release candidate for ~18 months is more than reason to not merge this, at least for the time being.

swwu:
Hey @jhpratt ! I wanted to follow up with some specifics about the limitations of bincode+serde, just for clarification's sake (although I agree that it doesn't make sense to merge this until `bincode2` is out of RC).

The main restrictions are around how bincode isn't a self-describing format, i.e. bincode can't serialize information about its own format. This means that it can't properly implement the `Deserializer::deserialize_any` method in serde. There's a list [here](https://docs.rs/bincode/2.0.0-rc.1/bincode/serde/index.html#known-issues) of all the things in serde that, for this reason, can't be implemented in bincode, but the short version is that anything that changes the structure of the produced serialization (skipping, flattening, enum tagging, etc) can't be done via bincode.

This is obviously annoying when working with `bincode` alongside, say, `serde_json`, since serde's default internal tagging for enums is fairly annoying to handle in a typescript/javascript frontend; this often means that you have to maintain different concrete structs (or manually implement a custom serializer/deserializer) if you want both ergonomic JSON output _and_ bincode on a particular struct. Bincode2 gets around this by implementing its own, non-serde-based `Encode` and `Decode` traits, that lets you use tagging/flattening/skipping to your heart's content for JSON, while still preserving bincode serializability.

Hope that helps!

jhpratt:
The only time `deserialize_any` is used in `time` is when implementing the visitor for `Option<T>`. It would be preferable if serde could handle container types (like `Option` and `Vec` natively, but that's not currently the case. Aside from that one instance, I don't see any reason it shouldn't work with bincode. None of the attributes mentioned on the linked page are used either.

swwu:
I think it would be less using `time` directly, but having a data structure that included one. For instance, you might have some type that looks like this:
```rust
#[derive(Serialize, Deserialize)]
#[serde(untagged)]
pub enum TimeOrElse {
  Time(Time),
  Else
}
```
which would serialize fine in `serde_json`, but not in bincode+serde, _even if_ `time` itself didn't run afoul of `deserialize_any`, because the `untagged` forces use of `deserialize_any`. In this case, it would be nice to do something like this instead:
```rust
#[derive(Serialize, Deserialize, bincode2::Encode, bincode2::Decode)]
#[serde(untagged)]
pub enum TimeOrElse {
  Time(Time),
  Else
}
```
Which would allow us to serialize/deserialize this struct using bincode as well as `serde_json`. But, of course, this only works if `Time: bincode2:Encode + bincode2: Decode`.

jhpratt:
That doesn't strike me as a particularly strong use case. As `time` works as expected on its own, I don't see the need to provide dedicated trait implementations. `bincode` could provide an implementation on the types directly or any crate can do the same via a wrapper struct.

PR #565, opened 2023-04-11
Add getters for ComponentRange

a1ien:
This methods is useful if you want to implement custom error printing.


a1ien:
Suggest then how to customize the error message. How to get what range is incorrect?
For example, in the current display implementation, you print the range as 1..=30, but this does not look good for the user interface.

PR #569, opened 2023-04-20
feat: add NAIVE_UNIX_EPOCH constant to PrimitiveDateTime

xoac:
I am not sure if this is acceptable in terms of interface `time` crate want to provide. I think `NAIVE` prefix is nice suggestion that developer is responsible for correctness of logic here and such constant can be useful.

### DIFF
--- a/time/src/date_time.rs
+++ b/time/src/date_time.rs
@@ -220,6 +220,12 @@ impl DateTime<offset_kind::None> {
         time: Time::MAX,
         offset: (),
     };
+
+    pub const NAIVE_UNIX_EPOCH: Self = Self {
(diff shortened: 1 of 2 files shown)


jhpratt:
"Unix epoch" fundamentally requires a time zone, as it measures time since 1970-01-01 0:00 **UTC**. If a user wishes to use the value without a time zone, it can be trivially constructed (at compile-time) using `datetime!(1970-01-01 0:00)`.

In the future, I intend on implementing additional functionality such that it's possible to have a statically-known offset with zero runtime overhead. However, this likely requires language features that are not near stabilization.

PR #580, opened 2023-05-18
Make time parsing more lenient

CaptainMaso:
I was trying to parse some values for time that I had previously been using chrono for, and found that the human readable parsing for time values required the subsecond field to be present.
I've added tests for this in the serde section, and then fixed them. All the tests still pass, but I guess this is probably a slightly controversial change as previously denied values are now accepted.

Please note that this change affects the Time, PrimitiveDateTime and OffsetDateTime parsing and deserialising, but it is not intended to affect the serialising/formatting.

The alternative is to require people to write their custom format_description, but this results in slightly annoying code to handle internal values when it's a pretty simple addition.

```rust
time::serde::format_description!(user_times, Time, "[hour]:[minute]:[second][optional [. [subsecond]]");

struct MyTime(time::Time);

impl<'de> serde_with::DeserializeAs<'de,time::Time> for MyTime {
    fn deserialize_as<D>(deserializer: D) -> std::result::Result<time::Time, D::Error>
    where
        D: serde::Deserializer<'de> {
        user_times::deserialize(deserializer)
    }
}

#[serde_as]
#[derive(Debug, Clone, serde::Deserialize)]
pub struct Config {
    #[serde_as(as = "Vec<MyTime>")]
    pub transmit_times_of_day: Vec<time::Time>,
}
```

### DIFF
--- a/tests/formatting.rs
+++ b/tests/formatting.rs
@@ -2,7 +2,7 @@ use std::io;
 
 use time::format_description::well_known::iso8601::{DateKind, OffsetPrecision, TimePrecision};
 use time::format_description::well_known::{iso8601, Iso8601, Rfc2822, Rfc3339};
-use time::format_description::{self, FormatItem, OwnedFormatItem};
+use time::format_description::{self, FormatItem, OwnedFormatItem, Component, modifier};
 use time::macros::{date, datetime, format_description as fd, offset, time};
 use time::{OffsetDateTime, Time};
 
@@ -271,6 +271,7 @@ fn format_time() -> time::Result<()> {
                 .format_into(&mut io::sink(), format_description)
                 .is_ok()
         );
+        
         assert_eq!(
             time!(13:02:03.456_789_012).format(&OwnedFormatItem::from(format_description))?,
             output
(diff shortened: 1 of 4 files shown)


jhpratt:
Can you add tests for the compact representation as well? Other than that, this PR looks good.

CaptainMaso:
It's probably worth a second review at this point. I've also added skipping formatting of optional values for some types of components under certain conditions.

The skippable components are seconds (when seconds past minute is 0), subseconds (when subseconds is 0), offset_hour (when the offset is UTC), offset_minute (when offset minutes past hour is 0), and offset_second (when offset seconds past minute is 0), and literals are always skippable. All other components are never skippable, as there isn't really a conceptual idea of when an hour item should be skipped, but it's pretty clear that when seconds or subseconds is 0, they can probably be skipped during formatting.

The logic is as follows:
 - Always format items unless they are wrapped in an `Optional` component.
 - Once wrapped in an Optional component, check if all sub-components are skippable (e.g. compound components require all internal components to be skippable to skip the whole component). 
 - If all sub-components are skippable, skip the formatting and emit 0 bytes written.
 - Otherwise, format as usual.
 
This means that literals that are wrapped in Optional are always skipped in formatting, so maybe we should include a modifier that makes it always formatted, but skippable in parsing? This use case is probably rare, as optional literals are usually paired with an actual parsing item that should control whether it is skipped or not.

jhpratt:
> I've also added skipping formatting of optional values for some types of components under certain conditions.

That's a significant change in behavior that is contrary to what documentation states. Optional items are _always_ present when formatting; they are only optional when parsing. I'd like to keep this PR to what was originally included, as making parsing more lenient and changing formatting behavior are two very different things.

CaptainMaso:
That's a fair assessment. I don't think that the work is inherently incorrect, but if the docs explicitly say the opposite then I guess that there's no real way to change that until 0.4 rolls around. I mostly did it because some tests assumed a round-trip conversion should be possible. 
I'll split the formatting work off into a separate PR and let that hang around until such a time that it might be usable. Until then I'll convert the tests I wrote to explicitly fail if the formatting changes, and get this one in a state to be merged.

jhpratt:
Closing as inactive.

PR #588, opened 2023-06-18
Make minutes, seconds optional when deserializing `UtcOffset`

MaxKingPor:
* Update `UTC_OFFSET_FORMAT` to accept more formats
* Added `visit_i64` and `visit_u64` for `Visitor<UtcOffset>` to accept numeric hours
* Unsafe expose `UtcOffset::__from_hms_unchecked`  as `UtcOffset::from_hms_unchecked`

### DIFF
--- a/tests/serde/mod.rs
+++ b/tests/serde/mod.rs
@@ -769,40 +769,40 @@ fn utc_offset_error() {
 #[test]
 fn utc_offset_partial() {
     assert_de_tokens_error::<Compact<UtcOffset>>(
-        &[Token::Tuple { len: 3 }, Token::TupleEnd],
+        &[Token::Tuple { len: 0 }, Token::TupleEnd],
         "expected offset hours",
     );
     assert_de_tokens_error::<Readable<UtcOffset>>(
-        &[Token::Tuple { len: 3 }, Token::TupleEnd],
+        &[Token::Tuple { len: 0 }, Token::TupleEnd],
         "expected offset hours",
     );
 
-    let value = offset!(+23:0:0);
+    let value = offset!(+23);
     assert_de_tokens::<Compact<UtcOffset>>(
         &value.compact(),
-        &[Token::Tuple { len: 3 }, Token::I8(23), Token::TupleEnd],
+        &[Token::Tuple { len: 1 }, Token::I8(23), Token::TupleEnd],
     );
-    let value = offset!(+23:0:0);
+    let value = offset!(+23);
     assert_de_tokens::<Readable<UtcOffset>>(
         &value.readable(),
-        &[Token::Tuple { len: 3 }, Token::I8(23), Token::TupleEnd],
+        &[Token::Tuple { len: 1 }, Token::I8(23), Token::TupleEnd],
     );
 
-    let value = offset!(+23:58:0);
+    let value = offset!(+23:58);
     assert_de_tokens::<Compact<UtcOffset>>(
         &value.compact(),
         &[
-            Token::Tuple { len: 3 },
+            Token::Tuple { len: 2 },
             Token::I8(23),
             Token::I8(58),
             Token::TupleEnd,
         ],
     );
-    let value = offset!(+23:58:0);
+    let value = offset!(+23:58);
     assert_de_tokens::<Readable<UtcOffset>>(
         &value.readable(),
         &[
-            Token::Tuple { len: 3 },
+            Token::Tuple { len: 2 },
             Token::I8(23),
             Token::I8(58),
             Token::TupleEnd,
--- a/tests/serde/mod.rs
+++ b/tests/serde/mod.rs
@@ -777,18 +777,18 @@ fn utc_offset_partial() {
         "expected offset hours",
     );
 
-    let value = UtcOffset::from_hms(23, 0, 0).unwrap();
+    let value = offset!(+23:0:0);
     assert_de_tokens::<Compact<UtcOffset>>(
         &value.compact(),
         &[Token::Tuple { len: 3 }, Token::I8(23), Token::TupleEnd],
     );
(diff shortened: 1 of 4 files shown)


jhpratt:
The only change here that I'm willing to make is having the minutes/seconds being optional.

Changing the padding to be `None` is a breaking change, and changing the sign to be optional is not ideal, as nearly every format for UTC offsets requires the sign. Adding `visit_i64` and `visit_u64` is functionally predicated on having the sign be optional.

With regard to having `UtcOffset::from_hms_unchecked`, there is no precedent for having this in `time`. Nothing can be constructed in an unchecked manner. If you want this currently, you can call `UtcOffset::from_hms` followed by `Result::unwrap_unchecked`.

MaxKingPor:
> 

Using `UtcOffset::from_hms` returns the `Result` type, which is not friendly to building global variables, and I want to maintain the unsafe `UtcOffset::from_hms_unchecked` directly by the developers themselves

jhpratt:
Global variables are an antipattern in Rust. If you're looking for a static value, however, you can use the `offset!` macro. I will not be providing a way to construct a value that can result in undefined behavior. It's simply unnecessary — there's the `.unwrap_unchecked()` method on results in addition to the macros. Unsafety is a last resort, not something to be reached for on a regular basis.

MaxKingPor:
> Global variables are an antipattern in Rust. If you're looking for a static value, however, you can use the `offset!` macro. I will not be providing a way to construct a value that can result in undefined behavior. It's simply unnecessary — there's the `.unwrap_unchecked()` method on results in addition to the macros. Unsafety is a last resort, not something to be reached for on a regular basis.

004AF45A
done

jhpratt:
CI is currently failing in a few ways. You'll almost certainly need to update the tests.

MaxKingPor:
> CI is currently failing in a few ways. You'll almost certainly need to update the tests.
> 
> Also, don't just do `.unwrap_or(0)`. This makes the assumption that if deserializing minutes fails, deserializing seconds will fail as well. While that is likely the case, it's not guaranteed.

done

jhpratt:
Thanks. I'll take a look at the PR in the coming days.

time/src/serde/visitor.rs
```diff
@@ -167,8 +167,17 @@ impl<'a> de::Visitor<'a> for Visitor<UtcOffset> {
...
+        let mut seconds = 0;
+
+        if let Ok(Some(min)) = seq.next_element() {
+            minutes = min;
+            if let Ok(Some(sec)) = seq.next_element() {
+                seconds = sec;
+            }
+        };
+        // let minutes = item!(seq, "offset minutes").unwrap_or(0);
+        // let seconds = item!(seq, "offset seconds").unwrap_or(0);
```
jhpratt, line 180: These lines can be removed.

tests/serde/mod.rs
```diff
@@ -770,36 +772,42 @@ fn utc_offset_partial() {
         &[Token::Tuple { len: 3 }, Token::TupleEnd],
         "expected offset hours",
     );
-    assert_de_tokens_error::<Compact<UtcOffset>>(
+    assert_de_tokens_error::<Readable<UtcOffset>>(
+        &[Token::Tuple { len: 3 }, Token::TupleEnd],
+        "expected offset hours",
+    );
+
+    let value = UtcOffset::from_hms(23, 0, 0).unwrap();
```
jhpratt, line 780: This (and other instances in the file) can be turned into a macro invocation.

jhpratt (changes requested):
Very minor changes. Overall, this looks good.

tests/serde/mod.rs
```diff
@@ -770,36 +772,42 @@ fn utc_offset_partial() {
...
+    assert_de_tokens::<Readable<UtcOffset>>(
+        &value.readable(),
+        &[Token::Tuple { len: 3 }, Token::I8(23), Token::TupleEnd],
+    );
+
+    let value = UtcOffset::from_hms(23, 58, 0).unwrap();
+    assert_de_tokens::<Compact<UtcOffset>>(
+        &value.compact(),
         &[
             Token::Tuple { len: 3 },
```
jhpratt, line 795: One other thing, actually. Shouldn't `len` here be 2, not 3? Why does this pass?

MaxKingPor: In 'Visitor<UtcOffset>', the default value of 0 is used when the number of minutes and seconds fails or none

time/src/serde/visitor.rs
MaxKingPor: done

tests/serde/mod.rs
MaxKingPor: done

tests/serde/mod.rs
jhpratt: Sure, but `len` is indicating the number of values present, which in this case is not 3.

tests/serde/mod.rs
jhpratt: No need to specify the zero values :)

tests/serde/mod.rs
MaxKingPor: ok I know, it's done

tests/serde/mod.rs
jhpratt: This is the only thing remaining, then I'll merge the PR :slightly_smiling_face:

MaxKingPor: Finish

jhpratt: No it's not? For example, `offset!(+23:0:0)` should be `offset!(+23)`.

MaxKingPor: Ok I see, has been revised

PR #596, opened 2023-07-08
Month and weekday arithmetic methods

### LINKED ISSUE #593: Advanced date arithmetic functionality
First of all, thank you for maintaining a crate that has proven very useful many times in my projects.

However, I think what this crate lacks is more advanced and high-level date arithmetic that can make writing code related to time calculations easier.

# Examples
- advance several months using one method or something simpler in general instead of having to manually calculate the target month and year
- calculate the first/nth occurence of weekday for a given month
- get the next/previous/nth weekday relative to a given point in time
- being able to get the time when a day/month/year starts (so basically rounding to given time units, although I see this has been mentioned in #524).

I believe this would be a great addition for this crate because operations on dates can be quite daunting because of all the inconsistencies in time units, like month lengths and leap years.

Also, this idea is quite general and would make for a very large contribution, so if you are interested, I could elaborate on some arithmetic features more specifically.

If this sounds like a good idea, I could try to contribute to it.

HubertK05:
This PR includes:
- nth_prev() method for Weekday,
- nth_prev() and nth_next() methods for Month,
- tests for methods listed above.

Closes #593.

### DIFF
--- a/tests/month.rs
+++ b/tests/month.rs
@@ -33,6 +33,74 @@ fn next() {
 }
 
 #[test]
+fn nth_next() {
+    assert_eq!(January.nth_next(0), January);
+    assert_eq!(January.nth_next(1), February);
+    assert_eq!(January.nth_next(2), March);
+    assert_eq!(January.nth_next(3), April);
+    assert_eq!(January.nth_next(4), May);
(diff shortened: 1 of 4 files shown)


jhpratt:
Looks great. Thank you!

PR #598, opened 2023-07-12
Change PartialEq, PartialOrd, Eq, Ord and Hash on Time for much more efficient versions

krtab:
This PR is split in three commits:

1. Add a benchmarking function that actually measure execution time of the comparison, as others are drowned in noise and/or optimized away IMO.
2. Adds the implementations themselves. :warning: This adds unsafe code.
3. Adds a more controversial optimization, see the message commit for pros and cons. This one can be removed without changing anything to the benchmarks and assembly below but may prevent code to be as efficient in more complex workloads.

## Benchmark

```
Time: vec_sort          time:   [96.691 µs 96.966 µs 97.262 µs]
                        change: [-54.795% -54.630% -54.470%] (p = 0.00 < 0.00)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe
```

## Generated assembly

```rust
pub fn gt(x: &Time, y: &Time) -> bool {
    x > y
}
```
**Before:**

```assembly
	movzx eax, byte ptr [rsi + 4]
	cmp byte ptr [rdi + 4], al
	jae .LBB357_2
	xor eax, eax
	ret

.LBB357_2:
	mov al, 1
	jne .LBB357_10
	movzx ecx, byte ptr [rsi + 5]
	cmp byte ptr [rdi + 5], cl
	jae .LBB357_5
	xor eax, eax
	ret

.LBB357_5:
	jne .LBB357_10
	movzx ecx, byte ptr [rsi + 6]
	cmp byte ptr [rdi + 6], cl
	jae .LBB357_8
	xor eax, eax
	ret

.LBB357_8:
	jne .LBB357_10
	mov eax, dword ptr [rdi]
	cmp eax, dword ptr [rsi]
	seta al

.LBB357_10:
	ret
```

**After:**

```assembly
        mov rax, qword ptr [rdi]
        cmp rax, qword ptr [rsi]
        seta al
        ret
```

### DIFF
--- a/benchmarks/time.rs
+++ b/benchmarks/time.rs
@@ -1,3 +1,5 @@
+use std::hint::black_box;
+
 use criterion::Bencher;
 use time::ext::{NumericalDuration, NumericalStdDuration};
 use time::macros::time;
@@ -206,6 +208,64 @@ setup_benchmark! {
         ben.iter(|| Time::MIDNIGHT < time!(0:00:01));
         ben.iter(|| time!(12:00) > time!(11:00));
         ben.iter(|| Time::MIDNIGHT == time!(0:00:00.000_000_001));
+    }
+
+    fn sort_align_8(ben: &mut Bencher<'_>) {
+        ben.iter_batched_ref(
+            || {
+                #[repr(C,align(8))]
+                struct Padder {
+                    arr: [Time;4096],
+                }
+                let mut res = Padder {
+                    arr: [Time::MIDNIGHT;4096]
+                };
+                let mut last = Time::MIDNIGHT;
+                let mut last_hour = 0;
+                for t in &mut res.arr {
+                    *t = last;
+                    t.replace_hour(last_hour).unwrap();
+                    last += 997.std_milliseconds();
+                    last_hour = (last_hour + 5) % 24;
+                }
+                res.arr.sort_unstable_by_key(|t|
+                    (t.nanosecond(),t.second(),t.minute(),t.hour())
+                );
+                res
+            },
+            |v| black_box(v).arr.sort_unstable(),
+            criterion::BatchSize::SmallInput
+        )
+    }
+
+    fn sort_align_4(ben: &mut Bencher<'_>) {
+        ben.iter_batched_ref(
+            || {
+                #[repr(C,align(8))]
+                struct Padder {
+                    pad: u32,
+                    arr: [Time;4096],
+                }
+                let mut res = Padder {
+                    pad: 0,
+                    arr: [Time::MIDNIGHT;4096]
+                };
+                let mut last = Time::MIDNIGHT;
+                let mut last_hour = 0;
+                for t in &mut res.arr {
+                    *t = last;
+                    t.replace_hour(last_hour).unwrap();
+                    last += 997.std_milliseconds();
+                    last_hour = (last_hour + 5) % 24;
+                }
+                res.arr.sort_unstable_by_key(|t|
+                    (t.nanosecond(),t.second(),t.minute(),t.hour())
+                );
+                res
+            },
+            |v| black_box(v).arr.sort_unstable(),
+            criterion::BatchSize::SmallInput
+        )
     }
     // endregion trait impls
 }
--- a/time/src/time.rs
+++ b/time/src/time.rs
@@ -29,21 +29,103 @@ pub(crate) enum Padding {
 /// (either positive or negative).
 ///
 /// When comparing two `Time`s, they are assumed to be in the same calendar date.
-#[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Clone, Copy, Eq)]
+#[repr(C)]
 pub struct Time {
-    #[allow(clippy::missing_docs_in_private_items)]
-    hour: u8,
-    #[allow(clippy::missing_docs_in_private_items)]
-    minute: u8,
-    #[allow(clippy::missing_docs_in_private_items)]
-    second: u8,
+    // The order of this struct's fields matter!
+    // Do not change them.
+
+    // Little endian version
+    #[cfg(target_endian = "little")]
     #[allow(clippy::missing_docs_in_private_items)]
     nanosecond: u32,
+    #[cfg(target_endian = "little")]
+    #[allow(clippy::missing_docs_in_private_items)]
+    second: u8,
+    #[cfg(target_endian = "little")]
+    #[allow(clippy::missing_docs_in_private_items)]
+    minute: u8,
+    #[cfg(target_endian = "little")]
+    #[allow(clippy::missing_docs_in_private_items)]
+    hour: u8,
+    #[cfg(target_endian = "little")]
     #[allow(clippy::missing_docs_in_private_items)]
     padding: Padding,
+
+    // Big endian version
+    #[cfg(target_endian = "big")]
+    #[allow(clippy::missing_docs_in_private_items)]
+    padding: Padding,
+    #[cfg(target_endian = "big")]
+    #[allow(clippy::missing_docs_in_private_items)]
+    hour: u8,
+    #[cfg(target_endian = "big")]
+    #[allow(clippy::missing_docs_in_private_items)]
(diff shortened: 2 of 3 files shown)


krtab:
Note that I only tested this on little-endian x86_64, so I didn't test the big endian code.

tests/meta.rs
```diff
@@ -115,7 +115,7 @@ fn size() {
     assert_size!(Date, 4, 8);
     assert_size!(Duration, 16, 16);
     assert_size!(OffsetDateTime, 16, 16);
-    assert_size!(PrimitiveDateTime, 12, 12);
+    assert_size!(PrimitiveDateTime, 16, 16);
```
jhpratt, line 118: This is far from ideal, and is directly caused by the difference in alignment. How much of a difference does the alignment make?

time/src/time.rs
```diff
@@ -29,21 +29,152 @@ pub(crate) enum Padding {
...
     minute: u8,
+    #[cfg(target_endian = "big")]
     #[allow(clippy::missing_docs_in_private_items)]
     second: u8,
+    #[cfg(target_endian = "big")]
     #[allow(clippy::missing_docs_in_private_items)]
     nanosecond: u32,
+    #[cfg(target_endian = "big")]
     #[allow(clippy::missing_docs_in_private_items)]
     padding: Padding,
```
jhpratt, line 72: The placement of this isn't correct for big endian platforms. It would cause the size of the struct to be larger.

time/src/time.rs
```diff
@@ -29,21 +29,152 @@ pub(crate) enum Padding {
...
+
+impl PartialEq for Time {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_u64().eq(&other.as_u64())
+    }
+
+    #[allow(clippy::partialeq_ne_impl)]
+    fn ne(&self, other: &Self) -> bool {
+        self.as_u64().ne(&other.as_u64())
+    }
```
jhpratt, line 89: Why implement various methods that have default implementations?

time/src/time.rs
```diff
@@ -29,21 +29,152 @@ pub(crate) enum Padding {
...
 impl Time {
+    /// Provides an u64 based representation **of the correct endianness**
+    ///
+    /// This representation can be used to do comparisons equality testing or hashing.
+    const fn as_u64(self) -> u64 {
+        // SAFETY:
+        // Transmuting to an u64 is always safe
+        // given that size is checked by transmute
+        // and alignment is irrelevant.
+        unsafe { core::mem::transmute(self) }
```
jhpratt, line 164: Why is `unsafe` necessary here? It should be possible to get the same result using `u64::from_le_bytes` or similar, I would imagine.

time/src/time.rs
krtab: My bad, fixed that!

time/src/time.rs
krtab: Because for some default methods, the compiler didn't manage to generate the optimal assembly that this hand implementation gives. Rather than manually checking all of them and relying on compiler optimization, I provided them.

time/src/time.rs
krtab: IIRC with `u64::from_le_bytes` the compiler didn't manage to see this was a transmute. And as above, I felt it was better to be more explicit rather than rely on compiler optimization.

krtab:
I've separated the PR in two.

In this one I don't force align to 8. On x64 it seems to be useless anyway

```
Time: sort              time:   [102.54 µs 103.97 µs 105.53 µs]
                        change: [-53.341% -52.613% -51.968%] (p = 0.00 < 0.00)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe

Time: sort_unaligned    time:   [103.88 µs 105.43 µs 107.15 µs]
                        change: [-52.691% -52.015% -51.322%] (p = 0.00 < 0.00)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
```

I've put the force align in another PR.

tests/meta.rs
krtab: Cf #599

time/src/time.rs
jhpratt: Do you remember which one (or multiple) didn't generate optimal assembly? It may be something that can be tackled in the compiler. Either way, it's less than ideal to implement method calls that are effectively redundant.

time/src/time.rs
jhpratt: Relying on compiler optimizations is fine. Some things in `time` already do. It's not as if the code will be broken if it's not as optimized.

time/src/time.rs
krtab: Well you were right to ask me to double check. I don't know if I did it differently this time, or I just did it wrong last time, but in any case, it seems to be ok without any unsafe code nor default impl now.

time/src/time.rs
```diff
@@ -29,21 +29,105 @@ pub(crate) enum Padding {
...
+    }
+}
+
+impl PartialEq for Time {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_u64().eq(&other.as_u64())
+    }
+}
+
+impl Eq for Time {}
```
jhpratt, line 85: I believe this can be derived even if `PartialEq` is manually implemented.

benchmarks/time.rs
```diff
@@ -207,5 +209,52 @@ setup_benchmark! {
...
+                    (t.nanosecond(),t.second(),t.minute(),t.hour())
+                );
+                res
+            },
+            |v| black_box(v).sort_unstable(),
+            criterion::BatchSize::SmallInput
+        )
+    }
+
+    fn sort_unaligned(ben: &mut Bencher<'_>) {
```
jhpratt, line 232: What is the difference between this and `sort`, in terms of what's being benchmarked?

benchmarks/time.rs
```diff
@@ -207,5 +209,52 @@ setup_benchmark! {
         ben.iter(|| time!(12:00) > time!(11:00));
         ben.iter(|| Time::MIDNIGHT == time!(0:00:00.000_000_001));
     }
+
+    fn sort(ben: &mut Bencher<'_>) {
+        ben.iter_batched_ref(
+            || {
+                let mut res = vec![];
+                let mut last = Time::MIDNIGHT;
+                for _ in 0..4096 {
```
jhpratt, line 218: With 4096 values, this only accounts for 68 minutes. Shouldn't the values be more spread out over the full range?

krtab: My logic was that it is enough to have code using all attributes of the struct, but I can make it larger if you prefer.

jhpratt: My thinking is that the CPU could take advantage of branch prediction to largely eliminate the check of the hour, assuming a different implementation. Benchmarks should be implementation-agnostic, after all.

krtab: Ok I will make it larger then!

jhpratt: And I'll note that it doesn't necessarily mean more iterations — having a larger value added each time would work as well.

time/src/time.rs
krtab: Done :heavy_check_mark:

benchmarks/time.rs
krtab: I've taken a slightly different route, because I felt like having a millisecond increment that is the closest prime to 1000 makes sense to go through every second and minute possible. I'm now incrementing hours independently by 5 (prime with 24) modulo 24. Hope that's fine.

jhpratt: Yeah, that works as well. I was just seeking something that would result in the values being more spread out, and that does the job.

benchmarks/time.rs
jhpratt: Gotcha. I'll take a look at the PR soon. Looks like CI will pass, so that's not a concern.

jhpratt (approved):

Issue #603, opened 2023-07-24
Parsing custom format without minutes, behavior inconsistent with seconds


GunnarMorrigan:
I want to be able to parse a custom format and let time fill in any other information as 0.

```Rust
    let format = time::macros::format_description!(
        "[year]-[month]-[day]_[hour]"
    );

    PrimitiveDateTime::parse("2023-05-12_19", &format)
```

Time crate currently returns 'the `Parsed` struct did not include enough information to construct the type'
This is technically right but it should fill in the blanks with 0. This is also actual behavior that the time crate already performs. If I add just the minute `"2023-05-12_19_00"` with format `"[year]-[month]-[day]_[hour]_[minute]"` then it works fine and sets seconds to 0.

The seconds are also missing from my format but why are they initialized to 0 and minutes do not do this? This is inconsistent behavior.

jhpratt:
In what situation is this format used? I have never encountered a time that only has the hours present (when not presented as a range, such as shop hours). The exception is for things like "1 PM", which is parsed successfully.

The reason I ask is because I only support parsing combinations that exist and are routinely used.

Zercerium:
just for reference #590
the [DWD](https://opendata.dwd.de/climate_environment/CDC/observations_germany/climate/hourly/precipitation/historical/) uses sometimes this format but there data shemas are really messed up sometimes 🙈 
```
STATIONS_ID;MESS_DATUM;QN_8;  R1;RS_IND;WRTR;eor
3;1995090100;    1;   0.0;   0;-999;eor
```

jhpratt:
As someone who has handled raw weather data before, I'm quite surprised that they opted to omit the minutes, as that's definitely relevant information. Regardless, that's sufficient for me. I'll add in support when I get a chance.

jhpratt:
Done in #604.

PR #604, opened 2023-07-27
Add support for only hours in time format with tests

GunnarMorrigan:
Per issue #603 add support for just hours in the time format. This allows leaving out the minutes when parsing a time or datetime object. 

If it is necessary to add more tests for specific cases or if the enhancement needs to be implemented through a different method than this simple change, please let me know.

The time macro was not adjusted as I think it does not makes sense to make `time!(12)` convert to 12:00 instead of 00:12. The user can simply write out the desired time.


GunnarMorrigan:
I will need to check and possibly update the documentation. Will add this

jhpratt:
Diff looks good. There's one thing I thought of while looking at it, but I don't want to put it on you. Merging as-is :slightly_smiling_face:

PR #615, opened 2023-08-21
Relax serde back after upstream issue resolution

### LINKED ISSUE #614: Relax back `serde` dependency
Hello I think the dependency can be now relaxed

https://github.com/serde-rs/serde/releases
https://github.com/serde-rs/serde/pull/2590

The precompiled binaries in the existing serde_derive versions have after long manual effort + process been ~reasonably audited / ~~reproducible~~ reverse-engineered (kind of) to some degrees so it may be perhaps fine to relax back to full range of possible versions - looking for input on this.

EDIT: Sadly cargo does not support relaxing two sets of ranges so would have to relax either 1) all or 2) provide cfg-gating

pinkforest:
Fixes #614 

I recommend relaxing back now that the issue has been resolved `>= 1.0.184` 

EDIT: Current approach is set to minimum 1.0.184

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ resolver = "2"
 
 [workspace.dependencies]
 time-core = { path = "time-core", version = "=0.1.1" }
-time-macros = { path = "time-macros", version = "=0.2.12" }
+time-macros = { path = "time-macros", version = "=0.2.13" }
 
 criterion = { version = "0.5.1", default-features = false }
 deranged = { version = "0.3.7", default-features = false }
@@ -16,8 +16,8 @@ num_threads = "0.1.2"
 quickcheck = { version = "1.0.3", default-features = false }
 quickcheck_macros = "1.0.0"
 rand = { version = "0.8.4", default-features = false }
-# <= 1.0.171 due to serde-rs/serde#2538
-serde = { version = ">= 1.0.126, <= 1.0.171", default-features = false }
+# ^1.0.184 due to serde-rs/serde#2538
+serde = { version = "1.0.184", default-features = false }
 serde_json = "1.0.68"
 serde_test = "1.0.126"
 trybuild = "1.0.68"
--- a/time/Cargo.toml
+++ b/time/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "time"
-version = "0.3.26"
+version = "0.3.27"
 authors = ["Jacob Pratt <open-source@jhpratt.dev>", "Time contributors"]
 edition = "2021"
 rust-version = "1.67.0"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog]. This project adheres to [Semantic Ver
 
 ---
 
+## 0.3.27 [2023-08-22]
+
+This sets the `serde` dependency requirement to `>= 1.0.184` where the binaries have been removed.
+
 ## 0.3.26 [2023-08-18]
 
 This release contains only a single change. `serde` is required to be a version prior to 1.0.171.
--- a/time-macros/Cargo.toml
+++ b/time-macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "time-macros"
-version = "0.2.12"
+version = "0.2.13"
 authors = ["Jacob Pratt <open-source@jhpratt.dev>", "Time contributors"]
 edition = "2021"
 rust-version = "1.67.0"


pinkforest:
Is there a way to create SemVer range like we did here: https://github.com/rustsec/advisory-db/pull/1417

We were able to use:

```toml
[versions]
patched = [">= 0.2.8, < 0.3.0-rc.1", ">= 0.3.0-rc.2"]
```

But with Cargo's dependency resolution I don't see / remembe a way how did you define multiple ranges.

Alternatively there is an approach to cfg-gate where by:
 - opt-in `RUSTFLAGS='--cfg precompiled="allow"'` -> opts-into full SemVer range
 - By default it only allows `>= 1.0.175`

Please let me know which way would be mergeable and I will adjust per preference.

Still looking a way to do multi-range scoping but I am not sure this was possible.

pinkforest:
@epage do we remember if there was a possibility in Cargo to do multi-range to exclude patch versions within the range like we did in https://github.com/rustsec/advisory-db/pull/1417 ?

time/Cargo.toml
```diff
@@ -50,10 +50,20 @@ deranged = { workspace = true }
...
 time-core = { workspace = true }
 time-macros = { workspace = true, optional = true }
 
+# Serde:
+# <= 1.0.171 due to serde-rs/serde#2538
+# >= 1.0.175 due to serde-rs/serde#2590
+# No way to support multiple version ranges e.g. >= 1.0.126
+# Opt-in Allow precompiled
+[target.'cfg(precompiled = "allow")'.dependencies]
+serde = { version = ">= 1.0.126", default-features = false, optional = true }
```
kayabaNerve, line 62: Alternatively, instead of `precompiled` there could be `older_serde` bound >= 1.0.126, <= 1.0.171`.

pinkforest: Yeah but that restriction would be opt-in ? will have to check with the maintainer but I suspect it has to be allow-in

kayabaNerve: By default, >= 185. By cfg, >= 126, < 172. Alternatively by cfg, >= 126.

One is for old serde, without precompiles. One is for old serde, including precompiles. One is for new serde.

That covers people who want to use old serde, yet don't want to allow precompiles.

pinkforest: How about -
```
cfg(time_serde_versions = "all") -> ">=1.0.126"
cfg(time_serde_versions = "pre-171") -> ">1.0.126, <= 1.0.171"
default -> ">= 1.0.185"
```

kayabaNerve: It's no longer explicit to being about precompiled, yet that would have the granularity I'm discussing. I'd let jhpratt decide what's optimal for time.

AlexTMjugador:
If `time` really wants to go with this route, can Cargo's platform-specific dependencies feature be used to make all these constraints exclusive to the `x86_64-unknown-linux-gnu` target? Precompiled binaries never were a concern for other platforms, as `serde` used traditional macro expansion instead.

AlexTMjugador:
> Is it guaranteed the blob will only execute on x86_64-unknown-linux-gnu and not any other supported target?

Yes, that is guaranteed at build-time via target flags, as you can see in the source code: https://github.com/serde-rs/serde/blob/bfcd44704f847ac5a9f3072e102e803b5ebbef31/precompiled/serde_derive/src/lib.rs#L18-L22

The binary is still downloaded as a part of the `serde_derive` crate files for other platforms, but it's unused.

kayabaNerve:
That wasn't my question.

I understand serde_derive will only execute it on x86_64-unknown-linux-gnu. I'm curious if it's only executable on x86_64-unknown-linux-gnu.

I don't personally believe time should have such platform-specific Cargo.toml validity.

AlexTMjugador:
> I understand serde_derive will only execute it on x86_64-unknown-linux-gnu. I'm curious if it's only executable on x86_64-unknown-linux-gnu.

If QEMU is installed and `binfmt_misc` set up for that, then yes, potentially any other Linux target can execute those binaries.

kayabaNerve:
I repeat I wouldn't be surprised if musl systems could run it (especially if they have gcompat). It should be able to if nothing is dynamically linked (which isn't still present on musl targets). FreeBSD also advertises Linux compatibility, and while it's optional, the question is now how many FreeBSD users have so opted in.

Unless it's known for sure it's not planting an in-the-future-executable blob, on any system, I'd personally still consider the binary blob a security risk.

jhpratt:
Given that Cargo doesn't have support for multiple ranges, my preferred solution at this point is to set the requirement to 1.0.184. Depending on the output of #611, it can be relaxed back to what it was before. The output of that seems nearly certain at this point, but there's still a few hours left.

With regard to `cfg`-gating other ranges, I think that'll cause more confusion than anything.

pinkforest:
So given `>= 1.0.184` is the preference here

**Risks**

This would break anyone who uses `<= 1.0.171` including when there is / are another crate/s who pinned.

I know several including `rsa` etc.

There would be two strategies

**Mitigation 1: Bump `time` to 0.4** 

So people who have either used `time = "0.3"` or `time = "0.3"` can bump it in controlled manner as releasing in 0.3 track would break anyone who uses serde with any dependency that has pinned to `<= 1.0.171` as time has so far done.

This approach would still regardless break anyone who has specified `time = "0"` which would have been illadvised regardless as it's not stable `1.0.0` yet.

**Mitigation 2: cfg approach**

I thought this would have been something I could get rest of the ecosystem motivated with to have unique mechanism to weed out desired dependencies depending on target environment requirements.

Personally I prefer 2) but just setting these out :)

I'll set the version to `>= 1.0.184` and I would recommend addressing the risk at least 1) mitigation if not 2)

Cargo.toml
```diff
@@ -16,8 +16,8 @@ num_threads = "0.1.2"
 quickcheck = { version = "1.0.3", default-features = false }
 quickcheck_macros = "1.0.0"
 rand = { version = "0.8.4", default-features = false }
-# <= 1.0.171 due to serde-rs/serde#2538
-serde = { version = ">= 1.0.126, <= 1.0.171", default-features = false }
+# >= 1.0.184 due to serde-rs/serde#2538
+serde = { version = ">= 1.0.184", default-features = false }
```
CryZe, line 20: You don't need the `>=`, that's the default.

pinkforest: Oh true, yes. Thank you. :heart:

Nemo157:
> This would break anyone who uses <= 1.0.171 elsewhere including when there is / are another crate/s who pinned.

It won't _break_ anything, it'll just not be possible to have those versions in the same buildgraph as the next version of time. As people relax/update their requirements too the ecosystem will naturally recover to the same state it was in, where the latest versions of everything are compatible.

pinkforest:
Yes, there can be time 0.3 and time 0.4 in buildgraph - but I thought oen can't have time 0.3.26 and time 0.3.27 same time as it's in the 0.3 range conflciting requirement. Unless one uses --locked and it will create build problems afaik ?

I'll double check this quick since that's how I understood it before and saw it happen - unless I'm dreaming :thinking:

Cargo.toml
Nemo157: (`>=` would actually allow be wrong as it allows updating to v2 etc., the default is `^` which means greater or equal within semver compatibility)

Nemo157:
I mean that just releasing this as 0.3.x won't break anything, there'll be a short time when it won't be possible to have the latest versions of everything in a build but once everyone has relaxed their requirements even will be good again. There's no mitigations required _unless_ some crates are not intending to allow use of 1.0.184+.

pinkforest:
Yes that is what I highlighted as a risk - the ones who pinned to up-range like time did but in any case maintainer will decide.

jhpratt:
0.4 wouldn't make sense, as it's not a breaking change. `--cfg` isn't ideal because of both the interaction with other crates and the non-testing in CI here (and I don't want to make CI times far worse).

The PR looks good as-is. Thank you @pinkforest and everyone else for their input! A release will be out shortly — CI takes about 25 minutes to validate everything for a proper release.

PR #620, opened 2023-09-16
chore(lib.rs): renamed no_coverage with coverate_attribute

hitenkoku:
Closed #619 

## What Changed

- replaced no_coverage with coverage_attribute due to no_coverage 

I would appreciate it if you could review when you have time.

PR #622, opened 2023-09-18
Fix llvm-cov build

ctz:
Another crack at #620.


jhpratt:
Could you remove the version bump? I do that at the same time as a release. Otherwise LGTM.

ctz:
> Could you remove the version bump? I do that at the same time as a release. Otherwise LGTM.

Done!

PR #628, opened 2023-10-08
Added closing markdown tag

Shagon94:
I was doing a exercism level and noticed this small thing. I also checked for more issues with `rg -i '/// ```' .` as to double-check if every opening tag has a closing one and from a quick view it looks good :)


jhpratt:
Thanks!

Issue #623, opened 2023-09-21
Missing panic condition on API docs


HeeillWang:
While I executed fuzzing, I faced some panics by `expect`. Though those are intentional crash, I think it's better to mention on docs as well. [Some of APIs](https://docs.rs/time/0.3.28/time/struct.Date.html#method.prev_occurrence) are already explicitly mention the panic condition.

I'll just list up the panic-condition-missing APIs, without reproduce condition as it's too obvious.

https://github.com/time-rs/time/blob/6248992b918638bc4f9e06e326c1fdace9c56fee/time/src/date.rs#L1304-L1307

https://github.com/time-rs/time/blob/6248992b918638bc4f9e06e326c1fdace9c56fee/time/src/duration.rs#L515-L535

https://github.com/time-rs/time/blob/6248992b918638bc4f9e06e326c1fdace9c56fee/time/src/duration.rs#L493-L513

https://github.com/time-rs/time/blob/6248992b918638bc4f9e06e326c1fdace9c56fee/time/src/duration.rs#L441-L453

jhpratt:
Doc improvements are always welcome! It's not a priority for me at the moment.

HeeillWang:
More cases : 

https://github.com/time-rs/time/blob/c96bb1a4474b9af1289edbdf34514fbfe95fa833/time/src/duration.rs#L1311-L1314

https://github.com/time-rs/time/blob/c96bb1a4474b9af1289edbdf34514fbfe95fa833/time/src/duration.rs#L1266-L1269

https://github.com/time-rs/time/blob/c96bb1a4474b9af1289edbdf34514fbfe95fa833/time/src/ext.rs#L239-L279

HeeillWang:
I'd like to defer adding doc for this at the moment, as I'm still running fuzzers, more cases may come out. 

However, anyone is welcomed to create pr for this.

jhpratt:
For panics that are deliberate, it's probably quickest to just check for `assert!` and `.expect` with ripgrep. Any other panics are likely unintentional — I can't think of any edge cases off the top of my head. `.unwrap` is not used anywhere, and this is enforced by clippy.

PR #638, opened 2023-11-29
Panic Documentation

gsquire:
This is an attempt to add some documentation for #623.

### DIFF
--- a/time/src/date_time.rs
+++ b/time/src/date_time.rs
@@ -1238,7 +1238,7 @@ impl From<DateTime<offset_kind::Fixed>> for SystemTime {
 impl From<js_sys::Date> for DateTime<offset_kind::Fixed> {
     /// # Panics
     ///
-    /// This may panic if the timestamp can not be represented by the return type.
+    /// This may panic if the timestamp can not be represented.
     fn from(js_date: js_sys::Date) -> Self {
         // get_time() returns milliseconds
         let timestamp_nanos = (js_date.get_time() * Nanosecond::per(Millisecond) as f64) as i128;
--- a/time/src/date_time.rs
+++ b/time/src/date_time.rs
@@ -1009,6 +1009,9 @@ impl<O: MaybeOffset> Hash for DateTime<O> {
 impl<O: MaybeOffset> Add<Duration> for DateTime<O> {
     type Output = Self;
 
+    /// # Panics
+    ///
+    /// This may panic if an overflow occurs.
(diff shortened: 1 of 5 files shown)


time/src/date_time.rs
```diff
@@ -1224,6 +1236,9 @@ impl From<DateTime<offset_kind::Fixed>> for SystemTime {
     feature = "wasm-bindgen"
 ))]
 impl From<js_sys::Date> for DateTime<offset_kind::Fixed> {
+    /// # Panics
+    ///
+    /// This may panic if the timestamp can not be represented by the return type.
```
jhpratt, line 1241:
```suggestion
    /// This may panic if the timestamp can not be represented.
```

time/src/duration.rs
```diff
@@ -1289,6 +1315,9 @@ impl Add<Duration> for StdDuration {
 impl_add_assign!(Duration: Self, StdDuration);
 
 impl AddAssign<Duration> for StdDuration {
+    /// # Panics
+    ///
+    /// This may panic if the resulting addition cannot be represented by the return type.
```
jhpratt, line 1320:
```suggestion
    /// This may panic if the resulting addition cannot be represented.
```

time/src/duration.rs
```diff
@@ -1336,6 +1374,9 @@ impl Sub<Duration> for StdDuration {
 impl_sub_assign!(Duration: Self, StdDuration);
 
 impl SubAssign<Duration> for StdDuration {
+    /// # Panics
+    ///
+    /// This may panic if the resulting subtraction can not be represented by the return type.
```
jhpratt, line 1379:
```suggestion
    /// This may panic if the resulting subtraction can not be represented.
```

jhpratt (commented):
Definitely a start! I only have a couple minor changes; it otherwise LGTM. The one for `Instant` actually made me think that it can be worked around, but as it's not yet, the comment is correct.

Also, don't worry about the coverage CI check. It's broken at the moment due to a bug in the Rust compiler.

time/src/duration.rs
gsquire: Updated as part of a single commit.

time/src/duration.rs
gsquire: Updated as part of a single commit.

gsquire:
Thanks for the review. I thought the code coverage was a separate issue so I ignored it for now. Let me know if you would like me to squash my commits.

PR #641, opened 2023-12-14
Add direct constructors for `OffsetDateTime`.

### LINKED ISSUE #640: OffsetDateTime lacks an obvious constructor
As evidenced by posts like this: https://www.reddit.com/r/rust/comments/18i8y39/learning_rust_my_experience_so_far_has_been_mixed/, as well as my own experience converting code from chrono to time, the lack of an obvious `OffsetDateTime::new` makes it more difficult to figure out how to work with the time crate.  

Some current workarounds:

Using macros.  This is how the docs tend to show creating an OffsetDateTime, but it only works if you have a literal datetime to plug in.  You can't use it with a list of integers representing different components, which is how you would want to, say, convert from a datetime gotten from another library or language.

Constructing a PrimitiveDateTime, and adding an offset to it.  This is an unsatisfying learning experience, because the user is trying to create an OffsetDateTime, so they aren't expecting the constructor to live on a different type.
```rust
PrimitiveDateTime::new(
    Date::from_calendar_date(2023, Month::December, 14)?, 
    Time::from_hms(12, 30, 59)?
).assume_utc()
```

Create an arbitrary OffsetDateTime, and replace all the components.  This also doesn't feel right, because you have to construct the wrong datetime to get to the right one, and you involve a syscall where it isn't necessary.  Also, you can get funky results but no error if you forget a component.  For instance, if you forget the nanoseconds, you could end up with two values that you expect to be equal that are not.

```rust
let dt1 = OffsetDateTime::now_utc()
    .replace_date(date)
    .replace_time(time)
    .replace_offset(offset);
let dt2 = OffsetDateTime::now_utc()
    .replace_year(year)?
    .replace_month(month)?
    .replace_day(day)?
    .replace_hour(hour)?
    .replace_minute(minute)?
    .replace_second(second)?
    .replace_nanosecond(nano)?
    .replace_offset(offset);
```

I propose adding a constructor:

```rust
fn new(date: Date, time: Time, offset: UtcOffset) -> OffsetDateTime {
     PrimitiveDateTime::new(date, time).assume_offset(offset)
}
```
and perhaps also a shortcut for utc, since that seems useful and common in time's api:

```rust
fn new_utc(date: Date, time: Time) -> OffsetDateTime {
     PrimitiveDateTime::new(date, time).assume_utc()
}
```

This still requires the user to know how to create a Date and a Time, but at least the references to those types are present in the `new` documentation, which is where they would expect it, and example code on the `OffsetDateTime` constuctors could show how to create them without requiring the user to navigate to the other types.

jcdyer:
Fixes #640

This PR adds new constructors to OffsetDateTime:
* `OffsetDateTime::new(date: Date, time: Time, offset: UtcOffset) -> Self`
* `OffsetDateTime::new_utc(date: Date, time: Time) -> Self`
* `OffsetDateTime::new_local(date: Date, time: Time) -> Result<Self, IndeterminateDateTime>`

Rationale for the PR can be found on the linked issue.

### DIFF
--- a/time/src/offset_date_time.rs
+++ b/time/src/offset_date_time.rs
@@ -71,65 +71,37 @@ impl OffsetDateTime {
     }
     // endregion now
 
-    /// Create a new `OffsetDateTime` with the given [[`Date`]], [`Time`], and [`UtcOffset`].
+    /// Create a new `OffsetDateTime` with the given [`Date`], [`Time`], and [`UtcOffset`].
     ///
     /// ```
     /// # use time::{Date, Month, OffsetDateTime, Time, UtcOffset};
     /// # use time_macros::datetime;
-    /// # fn main() -> Result<(), time::error::Error> {
-    /// let dt = OffsetDateTime::new(
+    /// let dt = OffsetDateTime::new_in_offset(
     ///     Date::from_calendar_date(2024, Month::January, 1)?,
     ///     Time::from_hms_nano(12, 59, 59, 500_000_000)?,
     ///     UtcOffset::from_hms(-5, 0, 0)?,
     /// );
     /// assert_eq!(dt, datetime!(2024-01-01 12:59:59.5 -5));
-    /// # Ok(())
-    /// # }
+    /// # Ok::<_, time::error::Error>(())
     /// ```
-    pub const fn new(date: Date, time: Time, offset: UtcOffset) -> Self {
+    pub const fn new_in_offset(date: Date, time: Time, offset: UtcOffset) -> Self {
         PrimitiveDateTime::new(date, time).assume_offset(offset)
     }
 
-    /// Create a new `OffsetDateTime` with the given [[`Date`]] and [`Time`] in the UTC timezone.
+    /// Create a new `OffsetDateTime` with the given [`Date`] and [`Time`] in the UTC timezone.
     ///
     /// ```
     /// # use time::{Date, Month, OffsetDateTime, Time};
     /// # use time_macros::datetime;
-    /// # fn main() -> Result<(), time::error::Error> {
     /// let dt = OffsetDateTime::new_utc(
     ///     Date::from_calendar_date(2024, Month::January, 1)?,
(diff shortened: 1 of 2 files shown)


jcdyer:
Can someone help me interpret the output of the coverage build, and understand what actions I should take?  To my eye, it looks like a failed snapshot in a part of the codebase I didn't touch.  Let me know if I'm missing something, and how I can resolve it.

time/src/offset_date_time.rs
```diff
@@ -71,6 +71,67 @@ impl OffsetDateTime {
     }
     // endregion now
 
+    /// Create a new `OffsetDateTime` with the given [[`Date`]], [`Time`], and [`UtcOffset`].
```
jhpratt, line 74:
```suggestion
    /// Create a new `OffsetDateTime` with the given [`Date`], [`Time`], and [`UtcOffset`].
```

in both places

time/src/offset_date_time.rs
```diff
@@ -71,6 +71,67 @@ impl OffsetDateTime {
...
+    ///
+    /// ```
+    /// # use time::{Date, Month, OffsetDateTime, Time, UtcOffset};
+    /// # use time_macros::datetime;
+    /// # fn main() -> Result<(), time::error::Error> {
```
jhpratt, line 79: This actually isn't necessary, as you can do `Ok::<_, time::Error>(())` at the end.

jhpratt (commented):
Don't worry about code coverage; the output is what I expect. It's currently slightly broken due to a bug in the compiler that I need to open an issue for. However, tests for these methods would be appreciated.

For the API itself, I would prefer not to have `now_local`. It would be quite error prone given that UTC offsets change throughout the year in most of the western world. `new_utc` is good as-is, though I'd like to see `new` renamed to `new_in_offset` for clarity.

jhpratt:
I just realized that this is only implemented in `offset_date_time.rs` and not in `date_time.rs`. Given that I'm reworking some of the internals in the latter at the moment, I'll handle that.

This PR looks good. Merging!

Issue #647, opened 2024-01-26
Cannot Deserialize timestamp with milliseconds


ogarcia:
My code is simple:
```rust
use serde::{Deserialize, Serialize};
use time::OffsetDateTime;

// Adds milliseconds support to timestamps
time::serde::format_description!(ts_milliseconds, OffsetDateTime, "[unix_timestamp precision:millisecond]");

#[derive(Debug, Deserialize)]
pub struct Profile {
    pub id: String,
    pub name: String,
    #[serde(with = "ts_milliseconds")]
    pub createdAt: OffsetDateTime,
    #[serde(with = "ts_milliseconds")]
    pub updatedAt: OffsetDateTime,
}
```

When I go to use the structure to deserialize from a JSON I get the following error.

```
invalid type: integer `1617099399918`, expected a(n) `OffsetDateTime` in the format \"[unix_timestamp precision:millisecond]\"
```

The truth is that I have run out of ideas, I understand that it should work as it is, but it does not. In the `cargo.toml` I have the following:
```toml
time = { version = "~0.3", features = ["formatting", "macros", "parsing", "serde"] }
```

jhpratt:
The value is expected to be a string, not an integer. Formatting and parsing is always done with strings.

ogarcia:
> The value is expected to be a string, not an integer. Formatting and parsing is always done with strings.

I understand then that it is impossible to deserialize a timestamp with milliseconds coming in an integer.

And you have not thought about including this possibility? It is not so rare nowadays to work with timestamps with milliseconds and the most normal thing is that they come as integers and not as strings.

In fact, I think it would be very interesting if it was even a module, something like `time::serde::timestamp_millis`. What do you think?

jhpratt:
I have previously rejected that (on multiple occasions) because I assumed it could be replicated using the `time::serde::format_description!` macro. Given that it cannot _and_ that `time::serde::timestamp` is serialized/deserialized as an integer, I would be fine with a PR adding it for milliseconds, microseconds, and nanoseconds.

PR #648, opened 2024-01-28
Adds support to serialize and deserialize timestamps with different resolutions

ogarcia:
As discussed in #647 this merge request adds the ability to serialize and deserialize timestamps with milliseconds, microseconds and nanoseconds.

For now I have only implemented the milliseconds because I want to see if it looks right to you @jhpratt if so then I do the rest.

The idea is that you can do the following:

```rust
use serde::{Deserialize, Serialize};
use time::serde::timestamp;
use time::OffsetDateTime;

#[derive(Serialize, Deserialize, Debug)]
struct TimestampSeconds {
    #[serde(with = "timestamp")]
    dt: OffsetDateTime,
}

#[derive(Serialize, Deserialize, Debug)]
struct TimestampMilliseconds {
    #[serde(with = "timestamp::milliseconds")]
    dt: OffsetDateTime,
}
```

### DIFF
--- a/tests/serde/timestamps.rs
+++ b/tests/serde/timestamps.rs
@@ -16,6 +16,18 @@ struct TestMilliseconds {
     dt: OffsetDateTime,
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+struct TestMicroseconds {
+    #[serde(with = "timestamp::microseconds")]
+    dt: OffsetDateTime,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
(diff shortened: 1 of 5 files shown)


jhpratt:
At quick glance this looks correct. My nits are minor at this point: I would split the tests for seconds/milliseconds/etc. into different functions, rather than just keeping it as one. I'd also indicate in the module documentation that it's the Unix timestamp _in milliseconds_ (as appropriate).

jhpratt:
LGTM. Thanks and good work!

PR #654, opened 2024-02-01
Allow unused_tuple lint under the new dead_code name

rillian:
As of at least rust 1.77 nightly, the compiler warns that the `unused_tuple_struct_field` lint has been subsumed under `dead_code`. Allow both lints in the relevant field here.

This still warns about the old name on nightly but helps with the severity of the warning for those building with a nightly toolchain and is less likely regress builds with future stable toolchains.

### DIFF
--- a/time-macros/src/lib.rs
+++ b/time-macros/src/lib.rs
@@ -68,7 +68,7 @@ enum FormatDescriptionVersion {
 #[cfg(any(feature = "formatting", feature = "parsing"))]
 enum VersionOrModuleName {
     Version(FormatDescriptionVersion),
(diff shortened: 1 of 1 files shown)


jhpratt:
Lints in `time` follow the latest stable, as that is what is used both in CI and when I am working on code. Formatting is the only thing that requires nightly, with documentation only using it for additional (non-essential) functionality.

It seems as though you think that users relying on `time` as a dependency will be shown this lint. That is thankfully not the case. Lints are only shown for the end crate and any path dependencies. Lints also don't stop working when renamed — they continue working as before, just with an additional lint (about the renaming) being emitted.

For these reasons, I'll be holding off until this reaches stable.

PR #655, opened 2024-02-02
Fix wasm32-unknown-unknown build

### LINKED ISSUE #652: Compilation error for "wasm-bindgen"
Compilation fails for `cargo build  --target wasm32-unknown-unknown --features "wasm-bindgen"`:
```log
error[E0425]: cannot find value `Millisecond` in this scope
    --> time/src/offset_date_time.rs:1600:69
     |
1600 |         let timestamp_nanos = (js_date.get_time() * Nanosecond::per(Millisecond) as f64) as i128;
     |                                                                     ^^^^^^^^^^^ not found in this scope
     |
help: consider importing one of these items
     |
3    + use crate::convert::Millisecond;
     |
3    + use time_core::convert::Millisecond;
     |

error[E0425]: cannot find value `Millisecond` in this scope
    --> time/src/offset_date_time.rs:1615:31
     |
1615 |             / Nanosecond::per(Millisecond).cast_signed().extend::<i128>())
     |                               ^^^^^^^^^^^ not found in this scope
     |
help: consider importing one of these items
     |
3    + use crate::convert::Millisecond;
     |
3    + use time_core::convert::Millisecond;
     |
```

This worked fine in v0.3.31, but now fails for both v0.3.32 and main

jedel1043:
This closes #652.

This also probably requires some CI modifications to test the `wasm_bindgen` builds, but that can be done in a follow-up PR.

### DIFF
--- a/time/src/offset_date_time.rs
+++ b/time/src/offset_date_time.rs
@@ -1596,6 +1596,7 @@ impl From<js_sys::Date> for OffsetDateTime {
     ///
     /// This may panic if the timestamp can not be represented.
     fn from(js_date: js_sys::Date) -> Self {
+        use time_core::convert::Millisecond;
         // get_time() returns milliseconds
         let timestamp_nanos = (js_date.get_time() * Nanosecond::per(Millisecond) as f64) as i128;
         Self::from_unix_timestamp_nanos(timestamp_nanos)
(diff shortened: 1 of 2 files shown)


time/src/offset_date_time.rs
```diff
@@ -1596,6 +1596,7 @@ impl From<js_sys::Date> for OffsetDateTime {
     ///
     /// This may panic if the timestamp can not be represented.
     fn from(js_date: js_sys::Date) -> Self {
+        use time_core::convert::Millisecond;
```
jhpratt, line 1599: I would prefer these imports to be at the top of the file. It would be reasonable to do `use crate::convert::*;` to avoid spelling everything out and dealing with gating it all correctly.

time/src/sys/local_offset_at/wasm_js.rs
```diff
@@ -1,3 +1,5 @@
+use num_conv::{CastSigned, Extend};
```
jhpratt, line 1:
```suggestion
use num_conv::prelude::*;
```

jhpratt (commented):
Coverage in CI is currently broken and there's no trivial way to add checking this to the powerset (which is run before each release). So don't worry about adding CI support at the moment.

jhpratt:
I'll get to this in the evening and cut a release at that point.

PR #656, opened 2024-02-02
fix unsigned to signed conversion bug

GregoryLakewood:
Casting `u32` into `i32` and then extending into `i64` does not hold the value when `val_u32 > i32::MAX`.
As explained in [Microsoft Documentation](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime#remarks) one should firstly construct `u64` from `FILETIME` and then perform arithmetics.

### DIFF
--- a/time/src/sys/local_offset_at/windows.rs
+++ b/time/src/sys/local_offset_at/windows.rs
@@ -59,10 +59,9 @@ fn systemtime_to_filetime(systime: &SystemTime) -> Option<FileTime> {
 /// Convert a `FILETIME` to an `i64`, representing a number of seconds.
 fn filetime_to_secs(filetime: &FileTime) -> i64 {
     /// FILETIME represents 100-nanosecond intervals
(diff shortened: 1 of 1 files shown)


jhpratt:
LGTM. As this won't affect (non-artificial) timestamps for 14 years, I won't be cutting a release specifically for this.

GregoryLakewood:
Thank you.
@jhpratt Could you look into this? I found this bug because I am getting `Err(IndeterminateOffset)` on my machine.
That breaks some Rust tools I use, specifically probe-rs.
From my understanding 31bit in lower FILETIME DWORD toggles every 2^31*100 nanoseconds what equals roughly 3 and half minutes.

jhpratt:
Ha, I wasn't thinking of the fact that it represents 100 nanosecond intervals. I was only off by a factor of ten million.

I'll be having another release later today.

jhpratt:
Released.

PR #657, opened 2024-02-02
Add must_use function attributes

cgzones:
Annotate several public funtions that either don't modify their original value or only compute a value, except for ones returning `Result`.

Avoid no-op code such as:

    let odt = OffsetDateTime::now_utc();
    odt.saturating_add(Duration::SECOND);

### DIFF
--- a/tests/date.rs
+++ b/tests/date.rs
@@ -1154,35 +1154,35 @@ fn nth_prev_occurrence_test() {
 #[test]
 #[should_panic]
 fn next_occurrence_overflow_test() {
-    date!(+999999 - 12 - 25).next_occurrence(Weekday::Saturday);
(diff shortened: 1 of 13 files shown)


jhpratt:
If there is a specific method you think could be reasonably interpreted as mutating the input that does not already have a `#[must_use]` attribute, please open an issue to discuss it. Adding it to everything is functionally pointless.

cgzones:
Is it though?
For example [std::time::Duration](https://doc.rust-lang.org/src/core/time.rs.html) uses the annotation on almost every interface (except for ones returning *Result*, since *Result* already comes with an annotation).

jhpratt:
And I disagree with using it that much. As I said, a method needs to be able to be reasonably interpreted as mutating the input. That is the criteria that I am using. Anything else is a non-starter.

PR #664, opened 2024-02-16
feat: added `Date::today()` method.

avhz:
Hi, thanks for the great library.

Just wondering if there could be a method added to get a `Date` with today's date.

`OffsetDateTime` has `.now_utc()` for example, which means I can do something like:

```rust
let today: Date = OffsetDateTime::now_utc().date();
```

But I thought it might be convenient for `Date` to have its own *now* method. e.g.

```rust
let today: Date = Date::today();
```

Happy to adjust the PR if needed of course.

### DIFF
--- a/time/src/date.rs
+++ b/time/src/date.rs
@@ -12,7 +12,6 @@ use num_conv::prelude::*;
 use powerfmt::ext::FormatterExt;
 use powerfmt::smart_display::{self, FormatterOptions, Metadata, SmartDisplay};
 
-use crate::convert::*;
 use crate::ext::DigitCount;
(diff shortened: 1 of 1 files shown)


jhpratt:
This is a deliberate omission that was present in v0.2 but removed in v0.3. Assuming a time zone is not an acceptable tradeoff.

avhz:
No worries. 

Could you clarify what you mean? I must be missing something. 
Where does the time zone data come into play if it returns a `Date`?

jhpratt:
It isn't the same day everywhere on earth.

avhz:
Ah I see what you mean. Cheers!

PR #665, opened 2024-02-18
Fix alignment of `Parsed` test

GuillaumeGomez:
This test is failing on latest rust nightly (since https://github.com/rust-lang/rust/pull/116672) and in `rustc_codegen_gcc` in the sync we're currently doing. Instead of having cfg, simpler to compare with `u128` type alignment directly, like that it will work in both nightly and the current version.

### DIFF
--- a/tests/meta.rs
+++ b/tests/meta.rs
@@ -26,12 +26,14 @@ use time::{
 #[test]
 fn alignment() {
     macro_rules! assert_alignment {
-        ($t:ty, $alignment:literal) => {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
As for the CI failing, seems like error messages got updated. Want me to include a commit to fix them?

jhpratt:
That one CI check has been failing for a little while now due to rust-lang/rust#119273.

Delegating alignment to what it is for i128 is definitely the smart thing to do. Thanks!

Issue #666, opened 2024-02-26
Optional leading zeros and parsing


softdevca:
The default Display implementation for `Time` and `OffsetDateTime` cannot reliably be parsed by a format description because the hour is not always two digits. 

```rust
 let format = format_description!("[hour]:[minute]:[second]");
 let time = Time::from_hms(1, 2, 3).expect("time");
 let parsed = Time::parse(&time.to_string(), &format).expect("parsed"); // Fails
```

The root issues seems to be that  in `parsing/component.rs` the following function requires the hour field be two digits, where the display implementation of `Time` does not pad to two digit hours:

``` rust
/// Parse the "hour" component of a `Time`.
pub(crate) fn parse_hour(input: &[u8], modifiers: modifier::Hour) -> Option<ParsedItem<'_, u8>> {
    exactly_n_digits_padded::<2, _>(modifiers.padding)(input)
}
```

I suggest that single digit hours be permitted.  A more general solution would be to enable format descriptions to indicate which fields may be single digits.

softdevca:
For a little more background:  A test was failing occasionally because code that was parsing the result of `OffsetDateTime::now_utc().to_string()` would start to fail around midnight UTC.

jhpratt:
Single digit hours can be parsed. You need to explicitly ask for that, as the default is to have them padded with zeroes. This is indicated [in the documentation](https://time-rs.github.io/book/api/format-description.html). More generally, you should use a format description for clarity rather than relying on the `Display` implementation if you are expecting a specific format. The output of `Display` is permitted to change without warning.

softdevca:
Submitted PR to the book to clarify padding.

PR #667, opened 2024-02-27
Clarify the format used by Display is may change

softdevca:
Additional documentation that the format of display may change and should not be relied upon, as indicated as comment in #666

### DIFF
--- a/time/src/date.rs
+++ b/time/src/date.rs
@@ -1415,6 +1415,7 @@ impl SmartDisplay for Date {
     }
 }
 
+/// The format returned by this implementation is not stable and must not be relied upon.
 impl fmt::Display for Date {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         SmartDisplay::fmt(self, f)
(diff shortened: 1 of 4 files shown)


jhpratt:
This is the case for any type implementing `Display` (in any crate) unless the documentation says it _can_ be relied upon. The stability guarantee is what is documented, and the output is not documented as stable.

softdevca:
I am unable to find any documentation the format of `fmt::Display` is unstable.  Not in the documentation for the Rust API (including `alloc` and `core), not in the Rust book or anywhere else.  It may be true that colloquially `Display`  cannot be relied upon but I respectfully believe it would be clearer if it was made explicit.  If for no other reason than to assist new users of the `time` crate .

jhpratt:
So long as the output is _correct_, any value is valid. The lack of a stability promise implies instability.  This is [semantic versioning](https://semver.org/), which deals with public API. In the case of `time` (and most Rust crates), this public API exists in the form of documentation. No where have I documented (and asserted the unchanging nature of) the output of `Display`, so it is simply not subject to semver.

softdevca:
Thank you for the clear and fair explanation.

PR #670, opened 2024-03-08
`Duration`: tighten documented ranges of sub-second values

### LINKED ISSUE #669: `subsec_{milli,micro,nano}seconds` may have over-inclusive documented ranges
Documentation of these functions promises that they will produce values in the ranges `-1_000..1_000`, `-1_000_000..1_000_000`, and `-1_000_000_000..1_000_000_000`, respectively, but I don't see how they could produce the lowest value in their stated bounds.

In terms of code, I'll use `subsec_nanoseconds` as an example:

https://github.com/time-rs/time/blob/7a6a800843e999807ddc4a317e95705f4ed82da5/time/src/duration.rs#L912-L923

`self.nanoseconds` is a [`Nanoseconds`](https://github.com/time-rs/time/blob/7a6a800843e999807ddc4a317e95705f4ed82da5/time/src/duration.rs#L29-L31), which is a [`RangedI32`](https://docs.rs/deranged/0.3.11/deranged/struct.RangedI32.html) - in particular, a `RangedI32<{ -(Nanosecond::per(Second) as i32 - 1) }, { Nanosecond::per(Second) as i32 - 1 }>`. `per` seems to be defined appropriately, so this gives us `RangedI32<-999_999_999, 999_999_999>`. Given that a `RangedI32<MIN, MAX>` promises to contain "[a]n `i32` that is known to be in the range `MIN..=MAX`", it's impossible for `-1_000_000_000` to inhabit `Nanoseconds`, and so to be produced by `subsec_nanoseconds`.

I believe the same logic can be extended to `subsec_microseconds` and `subsec_milliseconds`, since they defer to `subsec_nanoseconds`.

samcowger:
Fixes #669.

### DIFF
--- a/time/src/duration.rs
+++ b/time/src/duration.rs
@@ -857,7 +857,7 @@ impl Duration {
 
     /// Get the number of milliseconds past the number of whole seconds.
     ///
-    /// Always in the range `-1_000..1_000`.
+    /// Always in the range `-999..=999`.
     ///
     /// ```rust
     /// # use time::ext::NumericalDuration;
@@ -885,7 +885,7 @@ impl Duration {
 
     /// Get the number of microseconds past the number of whole seconds.
     ///
-    /// Always in the range `-1_000_000..1_000_000`.
+    /// Always in the range `-999_999..=999_999`.
     ///
     /// ```rust
     /// # use time::ext::NumericalDuration;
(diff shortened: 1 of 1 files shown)


jhpratt:
Thanks!

PR #671, opened 2024-03-16
Delete identity conversion in format_description::parse_owned

dtolnay:
We have `format_items: impl Iterator<Item = Result<Item<'_>, Error>>` and `items: Box<[Item<'_>]>`.

The `Into::into` on the deleted line is resolving to `impl<T> From<T> for T` with T=`Item` which is not load-bearing.

The `into` on the line after this is `impl From<Box<[Item<'_>]>> for OwnedFormatItem`.


jhpratt:
Not sure how I missed this when it was committed. Presumably I used it for debugging at some point. In any situation, thanks!

PR #682, opened 2024-05-25
Update libc:musl to support loongarch64

zhaixiaojuan:


jhpratt:
This serves no purpose. `Cargo.lock` is only used during the development of the crate. Any downstream user completely ignores it.

PR #686, opened 2024-06-08
Expose `Time::MIN` and `Time::MAX`

jrandolf:
This PR exposes the internal max/min values for `Time`. These are useful for building ranges.

### DIFF
--- a/time/src/time.rs
+++ b/time/src/time.rs
@@ -154,14 +154,22 @@ impl Time {
 
     /// The smallest value that can be represented by `Time`.
     ///
-    /// `00:00:00.0`
-    pub(crate) const MIN: Self =
+    /// ```rust
+    /// # use time::Time;
+    /// # use time_macros::time;
+    /// assert_eq!(Time::MIN, time!(00:00:00.0));
+    /// ```
+    pub const MIN: Self =
         Self::from_hms_nanos_ranged(Hours::MIN, Minutes::MIN, Seconds::MIN, Nanoseconds::MIN);
 
     /// The largest value that can be represented by `Time`.
     ///
-    /// `23:59:59.999_999_999`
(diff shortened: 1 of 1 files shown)


jhpratt:
Can you elaborate on your use case? Even within this repository the values are used in very few locations, all of which could easily be replaced without issue.

jun-sheaf:
> Can you elaborate on your use case? Even within this repository the values are used in very few locations, all of which could easily be replaced without issue.

We have a scheduling app that accepts missing start and end times. To make computations easier, we replace these bounds with actual values, specifically `Time::MIN` and `Time::MAX`.

jhpratt:
In that case, why wouldn't you want to be explicit about the exact value being added?

jun-sheaf:
> In that case, why wouldn't you want to be explicit about the exact value being added?

We are being explicit. By not specifying the range ends, you opt in to MIN as the minimum and MAX as the maximum. We could enforce using MIN instead of a missing lower bound, but for the upper bound, the problem is slightly more difficult as specifying an upper bound is dependent on the precision.

That said, we were able to workaround MIN/MAX by subtracting 1.nanosecond() from MIDNIGHT. If you're not convinced of this PR, maybe we can add some documentation about max bounds and how to achieve them (perhaps using the workaround just mentioned).

jun-sheaf:
@jhpratt Gentle ping, could you please take a look? If you feel this PR isn't useful, please close.

jhpratt:
I've decided that this would cause minimal confusion given that `Ord` is already implemented. However, `MIDNIGHT` already exists with identical semantics as `MIN`, so I'd like to avoid exposing that (I'll probably look into removing it). I would accept an update that makes `MAX` and only `MAX` public.

PR #692, opened 2024-07-07
Fix sub-second negative Duration serialization.

vriesk:
This addresses the problem that the Duration's sign will "disappear" both during serialization and deserialization of sub-second Durations.

### DIFF
--- a/time/src/parsing/parsable.rs
+++ b/time/src/parsing/parsable.rs
@@ -526,9 +526,7 @@ impl sealed::Sealed for Rfc3339 {
         //   readability, to specify a full-date and full-time separated by
         //   (say) a space character.
         // Specifically, rusqlite uses space separators.
-        let input = input
-            .get(1..)
-            .ok_or_else(|| InvalidComponent("separator"))?;
+        let input = input.get(1..).ok_or(InvalidComponent("separator"))?;
 
         let input = exactly_n_digits::<2, _>(input)
             .and_then(|item| item.consume_value(|value| parsed.set_hour_24(value)))
@@ -633,9 +631,7 @@ impl sealed::Sealed for Rfc3339 {
         //   readability, to specify a full-date and full-time separated by
         //   (say) a space character.
         // Specifically, rusqlite uses space separators.
-        let input = input
-            .get(1..)
-            .ok_or_else(|| InvalidComponent("separator"))?;
+        let input = input.get(1..).ok_or(InvalidComponent("separator"))?;
 
         let ParsedItem(input, hour) =
             exactly_n_digits::<2, _>(input).ok_or(InvalidComponent("hour"))?;


time/src/serde/visitor.rs
```diff
@@ -58,7 +58,10 @@ impl<'a> de::Visitor<'a> for Visitor<Duration> {
             de::Error::invalid_value(de::Unexpected::Str(nanoseconds), &"nanoseconds")
         })?;
 
-        if seconds < 0 {
+        if seconds < 0
+            // make sure sign does not disappear when seconds == 0
+            || (seconds == 0 && value.starts_with("-"))
```
vriesk, line 63: Alternatively, this whole condition could be just `value.starts_with("-")`, but I think this makes the intention clearer.

time/src/macros.rs
```diff
@@ -65,7 +65,7 @@ pub use time_macros::datetime;
 /// );
 /// # Ok::<_, time::Error>(())
 /// ```
-/// 
+///
```
jhpratt, line 68: You'll want to run `cargo +nightly fmt` to ensure changes like this don't happen.

time/src/serde/mod.rs
```diff
@@ -252,10 +252,18 @@ impl Serialize for Duration {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         #[cfg(feature = "serde-human-readable")]
         if serializer.is_human_readable() {
+            // Simply extracting sign and then calling `self.whole_seconds().abs()` will
+            // cause overflow on `i64::MIN` seconds, hence this kludge.
```
jhpratt, line 256: Using `.unsigned_abs()` avoids the potential overflow while also being quite performant.

time/src/serde/mod.rs
```diff
@@ -252,10 +252,18 @@ impl Serialize for Duration {
...
+            } else {
+                ""
+            };
+
             return serializer.collect_str(&format_args!(
-                "{}.{:>09}",
-                self.whole_seconds(),
-                self.subsec_nanoseconds().abs()
+                "{sign_fix}{seconds}.{nanoseconds:>09}",
+                seconds = self.whole_seconds(),
```
jhpratt, line 265: There's no need to name these, as they're used in the order they're provided.

time/src/serde/visitor.rs
jhpratt: Yeah, this is fine as-is.

time/src/serde/mod.rs
vriesk: The point of named format args is not only to allow reordering, but also making the format string more readable (and less mistake-prone).

time/src/macros.rs
vriesk: Done.

time/src/serde/mod.rs
vriesk: Oh, I somehow missed it. Done.

vriesk (commented):
PTAL

time/src/duration.rs
```diff
@@ -41,7 +41,8 @@ type Nanoseconds =
 pub struct Duration {
     /// Number of whole seconds.
     seconds: i64,
-    /// Number of nanoseconds within the second. The sign always matches the `seconds` field.
+    /// Number of nanoseconds within the second. The sign always matches the `seconds` field,
+    /// except when `seconds` is 0.
```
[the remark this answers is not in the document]
vriesk: Also, even if you take your mentioned definiton on the face value, this is still not true, because a specific sign does not really match lack of sign.

time/src/serde/mod.rs
jhpratt: Reordering isn't desirable here, as they _should_ be in that order. I'd prefer leaving them as-is (i.e. without explicit names). It's standard practice in my experience.

time/src/duration.rs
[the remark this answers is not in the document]
jhpratt: It's not a Rust thing so much as a number thing. However, Rust does have `is_positive` and `is_negative` methods for signed integers, both of which return false for zero. I'd be open to an "if any" type of wording, which would arguably be more correct. The wording as changed, though, I don't quite like.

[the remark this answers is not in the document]
vriesk: I believe this comment might have lead to the error that I'm fixing here in the first place - making the implementer think that they can just copy the `seconds`'s field sign and be done.

As you said, the sign of 0 is a "number thing" and it's completely arbitrary whether 0 is defined to have no sign, to have both positive and negative sign, or something yet else (e.g. distinct negative 0 and positive 0 that some arithmetics have).

Therefore, I'd really prefer to leave out any implicitly assumed defintion from here (that is required to understand what would "if any" mean), and just be specific that 0 is specifically excluded from the sign match.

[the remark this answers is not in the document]
jhpratt: Given that I'm the one that implemented both, that's certainly not the case (I naturally know the many invariants present in the crate). This is a bug that wouldn't have been prevented by an addition to documentation here. I've probably never checked that documentation a single time after it was written, as I've internalized the invariant. Unfortunately that doesn't prevent bugs, as I am human and sometimes forget things. I would just as easily forget to check documentation, so that wouldn't help me out there.

The only time I'm aware of zero being considered anything other than unsigned is in number theory (where it can be either or both positive and negative), calculus (where it's zero _from a certain direction_), and floating point numbers (where it can be signed outright). To say the obvious, none of those are in any way relevant here; the sign of an `i64` is unambiguous in this context. Positive if greater than zero, negative if less than, and nonexistent if equal.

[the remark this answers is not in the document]
vriesk: I must say I don't understand your position. My comment makes the most possible explicit statement without needing any assumptions.

Your "if any" proposal requires the reader to understand that neither number theory, nor calculus, nor signed 0 arithmetics like floating point numbers have are relevant here, and in addition understand that "no sign present" is implied by both `.is_positive()` and `.is_negative()` returning false (because the Rust docs don't say anywhere explicitly what sign presence is/is not). That is a much bigger mental effort, even if the incorrect doc was not a cause of this particular bug.

vriesk:
PTAL? :)

jhpratt:
This is still waiting on resolution to the changed comment.

vriesk:
OK. But I replied there a while ago. I can just remove the comment change
from the PR if you still disagree with me.

jhpratt:
That is what I had initially requested, yes.

vriesk:
> That is what I had initially requested, yes.

Done, PTAL.

time/src/duration.rs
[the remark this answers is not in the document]
vriesk: Reverted that change.

jhpratt:
Sorry, I did not mean to close this! It happened automatically (somehow) when force-pushing a rebase.

Due to a quirk of GitHub, any PR that is force-pushed after being closed cannot be re-opened. As such I'll be merging it manually, so as to avoid making you open a new PR. You're still credited as the author :slightly_smiling_face:

PR #697, opened 2024-07-29
Fixed compilation error using rust 1.80.0

soucosmo:


jhpratt:
This change does literally nothing.

Issue #696, opened 2024-07-29
Build failure under 1.80.0


havok4u:
I recently upgraded to 1.80.0 from 1.79.0 and ran my program and got the following when the crates were being pulled down, during a cargo run.  
```
error[E0282]: type annotations needed for `Box<_>`
  --> /home/xxxxx/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.34/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `time` (lib) due to 1 previous error
```
If I change back to 1.79.0 and re-run then the problem goes away.  

Thanks

jhpratt:
#681
#685
#689
#693
#695

Issue #693, opened 2024-07-25
error[E0282]: type annotations needed for `Box<_>`


gc-victor:
After updating Rust to version 1.8.0, the following error appears:

```
error[E0282]: type annotations needed for `Box<_>`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.34/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

For more information about this error, try `rustc --explain E0282`
```

the-emerald:
Updating `time` to 0.3.36 (from 0.3.34) fixed it for me!

Mouwrice:
> Updating `time` to 0.3.36 (from 0.3.34) fixed it for me!

Can confirm! Had the exact same situation and solution.

jhpratt:
This was previous raised in #681, #685, and #689.

liudonghua123:
I have the similar issues when build a tauri app on macos. It's strange that the tauri app build successfully on linux and macos, but not on macos.

I also noticed that it seems [`Compiling tao`](https://github.com/tauri-apps/tao) logging appears on macos building logs, but not on linux.

See https://github.com/liudonghua123/pxt-microbit-app/actions/runs/10517054766/job/29140512953.

Issue #689, opened 2024-07-03
Ferrocene compiler needs more information about the type of Box<_> at compile time.


benracine:
When attempting to compile this crate using the Ferrocene compiler, I see the following error in `/src/format_description/parse/mod.rs`. 

![image](https://github.com/time-rs/time/assets/798971/de308ef9-7c36-4f20-a6c1-9e0228372936)

jhpratt:
This isn't specific to Ferrocene, nor is it an issue at all. See #681 — it's already been fixed.

Issue #685, opened 2024-06-06
compile error: time-0.3.29/src/format_description/parse/mod.rs:83:9  "type must be known at this point"


phial3:
```text
error[E0282]: type annotations needed for `Box<_>`
  --> /Users/admin/.cargo/registry/src/rsproxy.cn-8f6827c7555bfaf8/time-0.3.29/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

   Compiling slab v0.4.9
   Compiling ppv-lite86 v0.2.17
For more information about this error, try `rustc --explain E0282`.
error: could not compile `time` (lib) due to 1 previous error
```

what should i do?

jhpratt:
See #681. Running `cargo update` should suffice.

hamirmahal:
If encountering this issue in CI on a GitHub Action runner, is the appropriate form of action to open an issue in https://github.com/actions/runner-images/?

PR #699, opened 2024-07-31
Added an explicit type to items variable

Cosiamo:
The app I've been building for months has randomly started throwing this error: "error[E0282]: type annotations needed for `Box<_>`". I don't use this Crate directly and haven't updated any that I do use in a couple of weeks, so I'm not sure why the Rust compiler doesn't like this syntax now.

image


jhpratt:
#681
#685
#689
#693
#695
#696
#697

PR #700, opened 2024-08-01
Allow spaces (and other things) as separators when parsing RFC3339

fasterthanlime:
I don't like it anymore than you probably do, but the spec says so!

(and rusqlite uses that)

cf. https://hachyderm.io/@fasterthanlime/112834567377457088

### DIFF
--- a/time/src/parsing/parsable.rs
+++ b/time/src/parsing/parsable.rs
@@ -520,9 +520,18 @@ impl sealed::Sealed for Rfc3339 {
         let input = exactly_n_digits::<2, _>(input)
             .and_then(|item| item.consume_value(|value| parsed.set_day(value)))
             .ok_or(InvalidComponent("day"))?;
-        let input = ascii_char_ignore_case::<b'T'>(input)
-            .ok_or(InvalidLiteral)?
-            .into_inner();
+
+        // RFC3339 allows any separator, not just `T`, not just `space`.
+        // cf. Section 5.6: Internet Date/Time Format:
+        //   NOTE: ISO 8601 defines date and time separated by "T".
+        //   Applications using this syntax may choose, for the sake of
+        //   readability, to specify a full-date and full-time separated by
+        //   (say) a space character.
+        // Specifically, rusqlite uses space separators.
+        let input = input
+            .get(1..)
+            .ok_or_else(|| InvalidComponent("separator"))?;
+
         let input = exactly_n_digits::<2, _>(input)
             .and_then(|item| item.consume_value(|value| parsed.set_hour_24(value)))
             .ok_or(InvalidComponent("hour"))?;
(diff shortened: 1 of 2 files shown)


jhpratt:
I definitely don't like this, but I also don't like [this from RFC2822](https://github.com/time-rs/time/blob/aca5c3652aff6fc7df3fbd8591c3e324d3a6ffb2/tests/parsing.rs#L100-L112).

It looks like tests are currently failing, as I explicitly tested for situations like this. The miri run will fail due to newer lints (it uses nightly unlike the rest of runs), so don't worry about that. With the tests updated, the change so far LGTM.

fasterthanlime:
It looks like codecov is unhappy because of two lines that it counts as being in the patch, but are not actually part of the lines I touched?

fasterthanlime:
Okay, the only failing tests is now miri, just like you said it would!

This is ready for review afaict, maybe a bit heavy on the comments — if you want to trim those a bit I don't mind! Feel free to adjust the PR before merging it (if the changes otherwise look good).

Note: I have not encountered any separators other than " " or "T" in the wild, so time _could_ be stricter than RFC3339 but.. I'm not sure what the rationale would be! Your pick.

fasterthanlime:
@jhpratt Anything I can do to help this land?

jhpratt:
@fasterthanlime Probably the only thing would be rebasing to get miri passing. That'll save me the ~30 seconds to do it myself. Aside from that, I've been extremely busy. I do intend on reviewing this quite soon though!

fasterthanlime:
Just rebased as requested! 30 seconds is 30 seconds 😎

time/src/parsing/parsable.rs
```diff
@@ -520,9 +520,20 @@ impl sealed::Sealed for Rfc3339 {
...
+        // cf. Section 5.6: Internet Date/Time Format:
+        //   NOTE: ISO 8601 defines date and time separated by "T".
+        //   Applications using this syntax may choose, for the sake of
+        //   readability, to specify a full-date and full-time separated by
+        //   (say) a space character.
+        // Specifically, rusqlite uses space separators.
+        let input = if input.is_empty() {
+            return Err(InvalidComponent("separator").into());
+        } else {
+            &input[1..]
```
jhpratt, line 534: Oddly enough, I neither documented this anywhere nor enabled the relevant clippy lints for the module, but generally speaking I try to avoid slicing even if it's obvious to a human that it cannot panic (as is the case here). While the compiler is likely able to elide the panic, the point at which it does so during compilation is past the point where it determines what is easily inlined. So this _could_ have an actual effect on performance even if it's elided.

This seems super nitpicky, but I would prefer to have it written in another manner so as to avoid slicing of the panicking variety. If you'd rather I do this, not an issue! I'll get this merged quite soon in any situation.

fasterthanlime: I refactored it to use `.get()` instead, which shouldn't panic and takes care of the empty case as well. it's even more readable!

jhpratt: That's even better than I expected it to look. The same thing exists on line 641 FYI.

fasterthanlime: Fixed line 641 and squashed all my commits into one

jhpratt:
Thanks again! Sorry about the delay.

PR #705, opened 2024-09-04
Make the date_time method on struct DateTimeOffset public

PolpOnline:
I don't know what were the motivations behind it being private (if any)


jhpratt:
I recommend you search before opening a pull request; this has been denied repeatedly.

PR #706, opened 2024-09-23
feat: Change to use `core::error::Error`

sorairolake:
Change to use [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html) instead of [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html). The MSRV is bumped up to 1.81.0 by this change.


jhpratt:
This violates the MSRV policy in the README. Additionally, I am very deliberate about bumping MSRV and track a number of features to ensure that I'm not increasing it for only a single thing.

sorairolake:
I think this will be mergeable once Rust 1.83 is released. So I think this can be kept as draft pull request.

jhpratt:
I do not intend to bump the MSRV the moment 1.83 is out. I will batch a number of changes together and do it all at once, not piecemeal. That's not to mention the countless changes made in this PR for unclear reasons.

sorairolake:
> That's not to mention the countless changes made in this PR for unclear reasons.

@jhpratt This is because the CI will not pass unless apply `cargo fmt` to unchanged files too. This is unrelated to this change, so I separated it out as df0491c.

PR #707, opened 2024-09-23
style: Remove unnecessary commas

sorairolake:
Looking at other codes, it seems better to eliminate these commas. So I'm going to remove these commas.

### DIFF
--- a/time/src/offset_date_time.rs
+++ b/time/src/offset_date_time.rs
@@ -74,7 +74,7 @@ impl OffsetDateTime {
     /// ```rust
     /// # use time::OffsetDateTime;
     /// # use time_macros::datetime;
-    /// assert_eq!(OffsetDateTime::UNIX_EPOCH, datetime!(1970-01-01 0:00 UTC),);
+    /// assert_eq!(OffsetDateTime::UNIX_EPOCH, datetime!(1970-01-01 0:00 UTC));
     /// ```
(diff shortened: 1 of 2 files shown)


jhpratt:
This likely isn't exhaustive, as I'm pretty sure I've run across more than just these in passing. Can you be sure that it is?

sorairolake:
@jhpratt This is applied to lines which match `rg -F "),)"`. So I cannot be sure that this is exhaustive.

First, I discovered a redundant comma in the following doctest:

https://github.com/time-rs/time/blob/2e68a36abc0d8cdc12638c0ed47b80cc9c07b7ab/time/src/offset_date_time.rs#L77

Next, I ran `rg -F "),)"` and discovered a similar redundant comma in the following line:

https://github.com/time-rs/time/blob/2e68a36abc0d8cdc12638c0ed47b80cc9c07b7ab/time/src/serde/mod.rs#L208

jhpratt:
If cleanup like this is being done, I'd like to be at least reasonably confident that it is exhaustive, lest we have a number of PRs substantially similar to this.

jhpratt:
@sorairolake What's the status of this?

sorairolake:
@jhpratt I'm not confident that I can do cleanups like this thoroughly.

jhpratt:
Closing for that reason. I'll probably end up digging through code myself with some more complicated regex in the near future.

PR #711, opened 2024-09-29
Update mod.rs [Bug Fix in the parse_owned Function, line 75-88]

a-zmuth:
# Overview

The change specifically addresses a bug fix in the `parse_owned` function and issues faced while installing the Anchor framework using AVM.

### Problem
The parse_owned function had compilation issues due to improper type handling in collecting format_items, leading to an error related to type conversion.

### Error Encountered
A compilation error indicated that OwnedFormatItem could not be created from Box<Vec<_>>.

```rust

error[E0282]: type annotations needed for `Box<_>`
  --> C:\Users\Username\.cargo\registry\src\index.crates.io-6f17d22bba15001f\time-0.3.29\src\format_description\parse\mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++
```
### Solution
The issue was fixed by explicitly specifying the type of collected format_items to Box<[format_item::Item<'_>]>, resolving the compilation error and improving type compatibility.


dennisorlando:
Hmm are you sure about the origin of this bug?
The provided logs refers to `time` version 0.3.29, which I just compiled successfully without reproducing such error.
Do you have a list of steps to reproduce this bug?

a-zmuth:
> Hmm are you sure about the origin of this bug? The provided logs refers to `time` version 0.3.29, which I just compiled successfully without reproducing such error. Do you have a list of steps to reproduce this bug?

The error specifically occurs when installing anchor with avm and the error comes from mod.rs in my time directory: I gave more details here: [Stackoverflow](https://solana.stackexchange.com/a/16689/25807)

To recreate the bug, first instal avm:

`cargo install --git https://github.com/coral-xyz/anchor avm --force`

After installation, you install anchor with avm:

`avm use latest` It will give a prompt to download the latest anchor version. 

I believe this error only occurs with avm others install directly with `cargo install --git https://github.com/coral-xyz/anchor --tag v0.30.1 anchor-cli`.  This gives rise to other errors too, but that can be resolved as shown in the StackOverflow link.

jhpratt:
Come on, this is beyond ridiculous. A basic search or even _looking at the pinned issues_ would show you that this was resolved months ago.

a-zmuth:
> Come on, this is beyond ridiculous. A basic search or even _looking at the pinned issues_ would show you that this was resolved months ago.

Doesn't seem so, since it keeps coming up.

jhpratt:
You're using a months-old version of `time`. No kidding it doesn't seem resolved. Common sense, please.

a-zmuth:
> You're using a months-old version of `time`. No kidding it doesn't seem resolved. Common sense, please.

My bad. It appears avm uses that older version as the default. 
Appreciate the clarification.

dennisorlando:
> > Come on, this is beyond ridiculous. A basic search or even _looking at the pinned issues_ would show you that this was resolved months ago.
> 
> Doesn't seem so, since it keeps coming up.

Perhaps https://github.com/coral-xyz/anchor would be a better place for this issue?
It might be just a "you should update time.rs" problem

PR #716, opened 2024-11-15
Update rstest and rstest_reuse; bump MSRV to 1.67.1

musicinmybrain:
These two updates aren’t inherently linked, but I thought it made sense to do them together.

I confirmed that `cargo test` still works in the `time/` directory (the only crate that currently uses these workspace crates).


jhpratt:
What is the benefit to bumping the dependency version?

musicinmybrain:
My motivation is that I am updating the [`rust-rstest` package](https://src.fedoraproject.org/rpms/rust-rstest) in Fedora, and that required [patching our `rust-time` package](https://src.fedoraproject.org/rpms/rust-time/pull-request/2) to allow the latest version. (We can use parallel-installable compat packages to maintain older versions of crate library packages in the distribution as needed, but these are a maintenance burden when they proliferate too much, so we try to only create compat packages when they are really necessary.) Since we have a [policy of offering patches upstream](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_all_patches_should_have_an_upstream_bug_link_or_comment), I opened this PR.

From an upstream perspective, I think it’s usually easier to update dev dependencies periodically in smallish steps rather than keep the same versions indefinitely until something breaks, but that’s a matter of opinion. I’m not aware of any concrete reason these dependencies need to be updated *right now*, so you can certainly just close this PR and move on if you don’t find it useful.

jhpratt:
I'm not necessarily opposed to this; I'll almost certainly end up merging it given the minimal impact. I'm asking all of this out of curiosity more than anything. What does Fedora need `rust-rstest` for, given that it's a dev dependency? Shouldn't crates be able to do all of this themselves?

decathorpe:
> What does Fedora need `rust-rstest` for, given that it's a dev dependency?

Because we - when possible - run tests for the crates we package.

Issue #721, opened 2025-01-09
Serde support for millisecond precision timestamps is broken by the lack of truncating i128 to i64.


JohnDowson:
Causes "i128 is not supported" runtime errors when used with `serde_json` and/or `bson`.

https://github.com/time-rs/time/blob/0a5021869081415571c648ba6c8dae33a7fe6883/time/src/serde/timestamp/milliseconds.rs#L20

https://github.com/serde-rs/json/issues/846

conradludgate:
Seems to be known https://github.com/time-rs/time/blob/0a5021869081415571c648ba6c8dae33a7fe6883/tests/serde/timestamps.rs#L79

jhpratt:
While the `serde_json` issue was closed, I don't see how this _isn't_ an issue on their end. A Unix timestamp in milliseconds, microseconds, or nanoseconds cannot be stored in a 64 bit integer. As far as I can tell this is working as intended; the serializer is where the issue lies.

JohnDowson:
Can it not? Storing millisecond offset from 01.01.1970 as an i64 should be totally fine for another 200 years, if I'm mathing this correctly.

jhpratt:
The supported range of years is ±9999 by default and ±999,999 with the `large-dates` feature enabled.

JohnDowson:
Though I do agree that not supporting 128 bit integers, and offering to jump straight to bigint as a solution is one of choices of all time.

jhpratt:
So…math-wise, 2⁶³ milliseconds larger than the 1 million supported. As such it was an oversight on my end when this was first implemented. This does _not_ apply to microseconds and nanoseconds, which would be 292,000 years and 292 years respectively.

I'd accept a pull request that duplicates the `time::serde::timestamp::milliseconds` module as `time::serde::timestamp::milliseconds_i64`, truncating the value (using `num-conv`'s methods) when serializing and requesting an `i64` when deserializing (and similarly calling `.extend()`).

jhpratt:
Closed by #722

PR #722, opened 2025-01-09
Truncate milliseconds to i64 before serializing.

JohnDowson:
A fix for #721


JohnDowson:
@jhpratt thank you for picking this up!

jhpratt:
Nah, this was on me. I did a rebase to get CI passing, then after a quick review I saw that deserialization was using `i128`. Changing that and using the relevant methods from `num-conv` is all that it took for me to be happy.

As simple of a PR as this is, thank you!

PR #748, opened 2025-08-13
Add `#[inline]` to `Time::from_hms` for performance

finnbear:
This makes [decoding `Time` in `bitcode`](https://github.com/SoftbearStudios/bitcode/pull/43) 6X faster.

It works since `bitcode` checks the range of integers up front, all at once, using vector instructions, and then uses `unreachable_unchecked` right before calling `Time::from_hms` to communicate that fact to `rustc`. This only works if `Time::from_hms` is inlinable.


jhpratt:
Incidentally I was planning on going through all functions in the coming days and weeks to (re)add `#[inline]` where appropriate. As such I'll _likely_ end up closing this at that point in time so as to avoid many separate commits for the same thing.

PR #779, opened 2026-05-07
typo?, in param of UtcDateTime::replace_minute / replace_second

BillGoldenWater:
stumbled on this when generating some glue code from source.

but not sure about it since it's like that for 2 years now.

searched issue/pr and found nothing related.


jhpratt:
Ha, I'd be shocked if this isn't me forgetting to rename things after copy-pasting because the functions are so similar. Good catch.

Issue #783, opened 2026-06-12
New version (0.3.48) makes my project stop compiling


nicopap:
Edit from @jhpratt:

Do not make "me too" comments. They add nothing of value and will only mean it takes longer for me to release a patch, as it's more I have to read. I am aware of the issue and the downstream breakage. A patch will be issued as soon as I am able.

Original issue follows.

---

Since updating the `time` crate from `0.3.47` to `0.3.48`, my crate stop compiling with the following error message:

```
error[E0119]: conflicting implementations of trait `From<format_description::parse::format_item::HourBase>` for type `<format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type`
   --> .../tracing-subscriber-0.3.23/src/filter/env/mod.rs:759:1
    |
759 | / impl<S> From<S> for EnvFilter
760 | | where
761 | |     S: AsRef<str>,
    | |__________________^
    |
    = note: conflicting implementation in crate `time`:
            - impl From<format_description::parse::format_item::HourBase> for <format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type;
    = note: upstream crates may add a new impl of trait `core::convert::AsRef<str>` for type `time::format_description::parse::format_item::HourBase` in future versions
```

Not sure if it is the `time` crate responsibility or `tracing_subscriber` 🙈

I'm not sure what's going on. `HourBase` is a private type. May be a compiler bug?

Tested with cargo v1.90.0, nightly 1.98.0 (2026-06-11) and nightly 1.96.0 ( 2026-03-19)

## Workaround

As mentioned in the thread, in your `Cargo.toml`, pin the `time` dependency:

```diff
-time = "0.3.47"
+time = "=0.3.47"
```

If time is a transitive dependency, then add the `+` line to your `Cargo.toml`. It's recommended to revert to unpinned dependency once the fix is in.

**Please don't do this if your crate is a library upon which 3rd parties rely** (🙏). It will likely result in duplicate `time` deps, and further compilation issues caused by incompatible same-crate types.

AntoineSebert:
Same issue with different toolchain:
```sh
rustup --version
rustup 1.29.0 (28d1352db 2026-03-05)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: the currently active `rustc` version is `rustc 1.96.0 (ac68faa20 2026-05-25)`
```

Workaround:
```
cargo update -p time --precise 0.3.47
```

[...]

PR #786, opened 2026-06-12
fix: resolve E0119 trait coherence error in rustc >= 1.95.0

itsarraj:
Replaced the use of the associated type `< as ModifierValue>::Type` with the concrete type via `target_ty!( ?)` in the generated `From` trait implementations inside `format_item.rs`.

This prevents trait coherence conflicts (E0119) in newer versions of the Rust compiler (1.95.0+), which are stricter about blanket implementations and associated types overlapping with downstream generic implementations (such as in `cookie` or `actix-web`).

### DIFF
--- a/time-macros/src/format_description/format_item.rs
+++ b/time-macros/src/format_description/format_item.rs
@@ -737,7 +737,7 @@ macro_rules! modifier {
                 type Type = target_ty!($name $($target_ty)?);
             }
 
-            impl From<$name> for <$name as ModifierValue>::Type {
+            impl From<$name> for target_ty!($name $($target_ty)?) {
                 #[inline]
                 fn from(modifier: $name) -> Self {
                     match modifier {
--- a/time/src/format_description/parse/format_item.rs
+++ b/time/src/format_description/parse/format_item.rs
@@ -700,7 +700,7 @@ macro_rules! modifier {
         }
 
         $(#[expect($expect_inner)])?
-        impl From<$name> for <$name as ModifierValue>::Type {
+        impl From<$name> for target_ty!($name $($target_ty)?) {
             #[inline]
             fn from(modifier: $name) -> Self {
                 match modifier {


itsarraj:
@jhpratt merge it

jhpratt:
Stop pinging people. It is obnoxious and will get you blocked.

mujeeb-kalwar:
How quickly can we get this released?

mujeeb-kalwar (approved):

agene0001:
any eta on when this will be merged. getting the same issue in my codebase

jhpratt:
The version in question has been yanked from crates.io. As such, there is no urgency. If I get more `Cargo.toml`s that were affected as mentioned in #783, that would speed up my analysis.

jplatte:
I think this fix is likely correct, as these impls are almost certainly triggering a rustc bug I've encountered before: https://github.com/rust-lang/rust/issues/85898

jhpratt:
Christ that's an old bug. Must not be terribly easy to fix if it's been around for over a decade.

Given that this already has a known compiler bug, I've got to agree that this is the fix. I'll get back around to it later before releasing.

jplatte:
This seems like the sort of thing that will be fixed, or at least the fix unblocked, by the new trait solver. Which is actually getting very close to passing a full crater run, AFAIK.

demurgos:
I've posted a minimal repro in the issue thread: https://github.com/time-rs/time/issues/783#issuecomment-4698179637.

Based on testing, I confirm that the issue is related to the usage of the associated type in `impl From<HourBase> for <HourBase as ModifierValue>::Type {...}`. The trait solver does not evaluate `<HourBase as ModifierValue>::Type` as `bool` which then causes the observed issue.

Since both `<HourBase as ModifierValue>::Type` and `bool` should be equivalent, this is definitely a Rustc bug (as linked above), and "evaluating" it manually by simply writing in the concrete type `bool` will fix this problem while keeping the same semantics. In general, the macro should expand to use public type for the modifier.

The macro is already defining the associated type using `target_ty!`:

```
        $(#[expect($expect_inner)])?
        impl ModifierValue for $name {
            type Type = target_ty!($name $($target_ty)?);
        }
```

So reusing it to define the impl target is the right thing to do, the fix in the PR is correct.

itsarraj:
Thanks all, ready when you are.

PR #791, opened 2026-06-26
Restore lexer depth on the unclosed-bracket error path

mahirhir:
Parsing a format description with an unclosed nested bracket (`[a[`, `[optional [`, `[hour[`) panics in debug builds via `debug_assert!(self.context().is_component())`, reached through `consume_whitespace`.

The opening bracket increments `self.depth`. On the branch where no closing bracket is found, the lexer returns `UnclosedOpeningBracket` without restoring `self.depth`. `consume_component` discards that error and re-derives the unclosed-bracket error itself, so the lexer keeps going with `depth` one too high; `context()` then reports the wrong context and the assert fails.

The fix decrements `self.depth` before returning, matching the entry state like the other early returns in this function. Release builds skip the assert, so the observable symptom is a debug-build panic; the underlying issue is the leftover depth on this error path.

### DIFF
--- a/time/src/format_description/parse/lexer_ast.rs
+++ b/time/src/format_description/parse/lexer_ast.rs
@@ -716,6 +716,13 @@ impl<'input, const VERSION: u8, const OWNED: bool> Lexer<'input, VERSION, OWNED>
 
         let Some(closing_bracket) = self.consume_closing_bracket() else {
             hint::cold_path();
+            // The opening bracket consumed above incremented `self.depth` but was never matched by
+            // a closing bracket. `consume_component` swallows this error and then re-derives the
+            // outer "unclosed bracket" error itself, so we must leave `depth` (and therefore
+            // `context()`) exactly as it was on entry; otherwise the subsequent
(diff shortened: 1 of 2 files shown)


jhpratt:
Please rewrite the description. LLM-generated writing is not acceptable, and frankly I have no idea what it even means. There is no such thing as a "native" parser. Let me know when the description is rewritten and I will take a look.

mahirhir:
Rewrote the description and title. You're right that "native parser" was meaningless; it's a debug_assert in the format-description lexer. The opening bracket increments depth and the unclosed-bracket error path didn't restore it, so a later context() assert trips on inputs like `[a[`. Happy to add a regression test covering those inputs if you'd want it in the same PR.

jhpratt:
Yes, please add a regression test for the inputs. They can go in the `parse_format_description.rs` test file. Also, if you could ensure that an LLM is not listed as co-author in the commit, that would be preferred.

jhpratt:
Looks good; thanks!

Issue #793, opened 2026-06-30
Breaking change in time v0.3.52 compared to v0.3.51


TobiasDeBruijn:
After running `cargo update` just now, I ran into a compile error coming from a dependency `cookie`, which depends on `time`:

Pin time to the newest version, specifically:
```
$ cargo update  -p time --precise 0.3.52
    Updating crates.io index
    Updating time v0.3.51 -> v0.3.52
    Updating time-macros v0.2.30 -> v0.2.31
note: pass `--verbose` to see 4 unchanged dependencies behind latest
```
```
error[E0061]: this method takes 2 arguments but 1 argument was supplied
   --> /home/tobias/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cookie-0.16.2/src/parse.rs:263:27
    |
263 |     let mut date = format.parse(s.as_bytes())?;
    |                           ^^^^^-------------- argument #2 of type `Option<Parsed>` is missing
    |
note: method defined here
   --> /home/tobias/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/time-0.3.52/src/parsing/parsable.rs:62:12
    |
 62 |         fn parse(&self, input: &[u8], defaults: Option<Parsed>) -> Result<Parsed, error::Parse> {
    |            ^^^^^
help: provide the argument
    |
263 |     let mut date = format.parse(s.as_bytes(), /* Option<Parsed> */)?;
    |                                             ++++++++++++++++++++++

    Checking jsonwebtoken v10.4.0
For more information about this error, try `rustc --explain E0061`.
error: could not compile `cookie` (lib) due to 1 previous error
```

Downgrading to time v0.3.51 resolves the issue:
```
 $ cargo update  -p time --precise 0.3.51
    Updating crates.io index
 Downgrading time v0.3.52 -> v0.3.51
 Downgrading time-macros v0.2.31 -> v0.2.30 (available: v0.2.31)
note: pass `--verbose` to see 4 unchanged dependencies behind latest
```
```
$ cargo c
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
```

jhpratt:
This is due to `cookie` relying on an internal API. Please report the issue there, as that's where the issue lies. I have _never_ documented the existence of that method, nor have I ever guaranteed anything regarding semver on that front.

Powerbyte7:
Would it be possible to add unnamable types to the method signatures to prevent them from being called by another crate in the future?

```rs
[...]

PR #797, opened 2026-06-30
Seal Parsable trait methods

Powerbyte7:
This PR prevents the issue in #793 from reoccurring in the future, by embedding an unnameable type in the trait method signatures. The logic is analogous to what can be seen in the following code:

```rs
mod outer_sealed {
    mod inner_sealed {
        pub struct Token;
        pub struct Seal<T>(T);
    }
    
    pub(self) use inner_sealed::Token;
    
    pub trait SealedTrait {
        fn method(&self, _: inner_sealed::Token);
        fn method_with_return(&self) -> inner_sealed::Seal<u32>;
    }
}

use outer_sealed::SealedTrait; // The trait remains publicly usable
// use outer_sealed::Token; // This causes an error, which is what we want
```

For more info on this practice, see the following:
https://rust-lang.github.io/api-guidelines/future-proofing.html#newtypes-encapsulate-implementation-details-c-newtype-hide
https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/

I have tested this PR by running `cargo test --all-features`, and also also by checking that the build fails when importing the unnameable type. This PR can be considered breaking in the same way the changes to #793 were, but will improve future API stability.

### DIFF
--- a/time/src/date.rs
+++ b/time/src/date.rs
@@ -1478,7 +1478,7 @@ impl Date {
         input: &str,
         description: &(impl Parsable + ?Sized),
     ) -> Result<Self, error::Parse> {
-        description.parse_date(input.as_bytes(), None)
+        description.parse_date(input.as_bytes(), None, crate::parsing::SealedToken)
     }
 
     /// Parse a `Date` from the input using the provided [format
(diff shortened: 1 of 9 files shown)


brianmay (approved):

jhpratt:
Thanks for the quick PR! I actually ended up going a bit farther, using a similar approach for the formatting trait as well. I'm not sure if it was technically necessary, but there's no harm in adding a ZST as a parameter. Given the larger scope and timeliness of the issue, I'm doing this myself. Sorry!

PR #807, opened 2026-08-27
Fix deprecation message: MonthNumeric -> MonthNumerical

d-e-s-o:
Deprecation messages for Component::Month mention MonthNumeric, but there is no such type. It seems to be called MonthNumerical instead. Adjust messages.


jhpratt:
Thanks

jhpratt:
Looks like there's some issues with CI at the moment. It's upstream, as I've only ever encountered this error with other projects (namely the rust-lang/rust repo). I'll retry again in a bit.

jhpratt:
Actually it was my fault :) merging manually so you don't have to rebase.

PR #808, opened 2026-08-27
Add direct link to documentation

d-e-s-o:
Make the URL to the documentation of the crate known in the manifest so that it will be easier to click through to it from a `crates.io` search.


jhpratt:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-documentation-field

> If no URL is specified in the manifest file, [crates.io](https://crates.io/) will automatically link your crate to the corresponding [docs.rs](https://docs.rs/) page when the documentation has been built and is available

d-e-s-o:
I don't know what that's supposed to mean, but if you check https://crates.io/search?q=time there is no direct link.

jhpratt:
File an issue for that, then? I am 100% certain the lack of a field is correct here.

PR #809, opened 2026-09-02
Fix alignment tests on 32-bit architectures

michel-slm:
On i686, for instance, 64-bit integers are 32-bit aligned

See

https://koji.fedoraproject.org/koji/taskinfo?taskID=149783770 https://kojipkgs.fedoraproject.org//work/tasks/4350/149784350/build.log

```
---- meta::alignment::case_02 stdout ----
thread 'meta::alignment::case_02' (3516) panicked at tests/integration/meta.rs:91:5:
assertion `left == right` failed: alignment of `time::signed_duration::SignedDuration` was 8
  left: 4
 right: 8
```


jhpratt:
Can this be extracted into a `const` to make it more readable? As-is, it's not the best. You'll also want to rebase so that CI can run correctly — I had to make a minor change.

